PR #24, opened 2026-04-11
fix: remove buggy eval schema

Perry2004:


reacher-z:
@Perry2004 please double-check if you removed all

--add-reviewer Perry2004

PR #124, opened 2026-05-04
refactor: a better package structure for building and releasing #123

### LINKED ISSUE #4: Release lite version test suite
It would be great to have a 30-50 task lite version of the benchmark.

### LINKED ISSUE #67: Harden bundled data against symlink-unaware build tools
## Observation

\`src/clawbench/data/\` contains 7 symlinks pointing out of \`src/\`:

\`\`\`
src/clawbench/data/test-cases       -> ../../../test-cases
src/clawbench/data/tools            -> ../../../tools
src/clawbench/data/extension-server -> ../../../extension-server
src/clawbench/data/chrome-extension -> ../../../chrome-extension
src/clawbench/data/shared           -> ../../../shared
src/clawbench/data/models           -> ../../../models
src/clawbench/data/eval             -> ../../../eval
\`\`\`

\`hatchling\` follows symlinks during \`pip wheel .\` today — I verified the built wheel \`dist/claw_bench-0.1.0-py3-none-any.whl\` contains all 153 \`task.json\` files, the full Dockerfile set, the chrome-extension, etc. So **the current release pipeline works**.

But the fallback posture is fragile. Any of these produces an empty or partial wheel:

- \`tar -c\` without \`-h\` / \`--dereference\` when archiving source for a mirror build
- \`git archive\` (does not follow symlinks by default)
- \`cp -r\` without \`-L\` into a CI staging dir
- A future hatchling plugin or PEP 517 frontend change in symlink semantics
- \`pip install -e .\` on Windows (symlinks work only with dev-mode privileges)

## Proposed fix options

1. **Copy-on-build hook** — \`[tool.hatch.build.hooks.custom]\` that materializes \`src/clawbench/data/\` from the repo root before packaging, deleting it after. Keeps the dev-mode symlinks (useful for live-reload) but guarantees wheel content regardless of tool behavior.
2. **\`force-include\`** — \`[tool.hatch.build.targets.wheel.force-include]\` entries that map repo-root dirs directly into \`clawbench/data/\` at wheel-build time. Drop the symlinks entirely.
3. **CI assertion** — add a release.yml step that runs \`unzip -l dist/*.whl | grep -c task.json\` and fails if <150. Catches breakage post-hoc but is cheaper than a packaging refactor.

Option 2 is the most robust; option 3 is the smallest patch.

## Priority

Medium — not broken today, but a silent packaging regression would ship an empty \`clawbench cases\` to every new user. The failure mode is invisible in local dev because the symlinks still resolve.

### LINKED ISSUE #118: DOCS: update v2 update

### LINKED ISSUE #119: FEAT: multi task suite selection

### LINKED ISSUE #123: REFACTOR: structure change for building & publishing
This is required to support the release CD workflow.

Perry2004:


Perry2004:
This refactoring PR closes: #123 #118 #4 #119 #67

PR #140, opened 2026-05-09
feat: add --resume flag to batch runner to skip completed jobs

erenup:
Pass --resume <BATCH_DIR> to reuse a previous batch's output directory and skip any (case x model) whose batch-logs/<case>-<model>.log already exists, so an interrupted batch can be picked up without redoing work.

## Test plan

- [ ]

### DIFF
--- a/src/clawbench/runner/batch.py
+++ b/src/clawbench/runner/batch.py
@@ -580,9 +580,28 @@ async def async_main(args: argparse.Namespace) -> int:
 
     _run_mod.docker_build(args.harness)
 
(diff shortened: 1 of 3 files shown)


reacher-z:
Hi @erenup — your PR itself looks clean (the diff only touches `batch.py` with the `--resume` logic). The CI failures here are caused by an unrelated pre-existing format issue on `main`: `judge.py` and `run.py` (introduced in #139) currently fail `ruff format --check`, which makes `static-check` fail on every open PR including this one.

Just opened #141 to format those two files. Once that's merged, please re-run CI on this PR (or rebase) and the checks should go green. No changes needed from your side.

Thanks for the contribution!

PR #154, opened 2026-05-12
fix: resolve #66 — Add unit-test coverage for run.py, batch.py, and doctor.py

### LINKED ISSUE #66: Add unit-test coverage for run.py, batch.py, and doctor.py
## Gap

Three core modules have zero unit tests:

| file | LOC | coverage |
|---|---|---|
| \`src/clawbench/run.py\` | 1009 | none |
| \`src/clawbench/batch.py\` | 628 | none |
| \`src/clawbench/doctor.py\` | 194 | none |

\`tests/\` currently ships \`test_cli.py\`, \`test_engine.py\`, \`test_image.py\`, \`test_paths.py\` — but not these three, which together are ~1.8k lines of the hottest code paths.

## Concrete bugs this would catch

- **run.py**: \`_load_runtime_env\` precedence (env → ./.env → user config → wheel default). A precedence regression silently leaks or hides keys. One test per source wins.
- **run.py**: result-JSON serialization for the HF upload path — malformed JSON breaks \`hf_upload\` only at the end of a long run.
- **batch.py**: case-range parsing (\`--case-range 1-50\`), glob filtering for \`--models\`/\`--cases\`, \`--all-models\` + \`--models\` precedence. Easy off-by-one territory.
- **batch.py**: concurrency respecting \`--stagger-delay\` and \`--max-concurrent\` under a mocked subprocess fan-out.
- **doctor.py**: each \`check_*\` returns correct status under its known-good and known-bad fixtures. Catches the classic "doctor says all green while models.yaml is actually broken" footgun.

## Scope

- 6–8 tests for run.py (focus on \`_load_runtime_env\`, arg parsing, result parsing)
- 5–7 tests for batch.py (case-range, glob filtering, dry-run matrix)
- 5–7 tests for doctor.py (one per check, with mocked engine + filesystem fixtures)
- No container runs in CI — everything at the subprocess/filesystem mock boundary

## Priority

Medium. These modules are user-facing (\`clawbench run\`, \`clawbench batch\`, \`clawbench doctor\`) and the CI today would let a regression ship.

SlncTrZ:
## Summary

fix: resolve #66 — Add unit-test coverage for run.py, batch.py, and doctor.py

## Problem

**Severity**: `Low` | **File**: `tests/conftest.py`

To avoid duplication across `test_run.py`, `test_batch.py`, and `test_doctor.py`, we introduce a `conftest.py` that provides reusable fixtures. These include a temporary directory fixture, an environment variable fixture (using `monkeypatch`), a fixture to mock `subprocess.run` with configurable return values, and a fixture to create dummy `models.yaml` or `.env` files. This improves test maintainability and keeps individual test files focused on logic.

## Solution

Create `tests/conftest.py` (if not already present; if present, append to it). Include:

## Changes

- `tests/conftest.py` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced"

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,34 @@
+import pytest
+from pathlib import Path
+import subprocess
+from unittest.mock import Mock
+
+@pytest.fixture
+def temp_dir(tmp_path):
+    return tmp_path
+
+@pytest.fixture
+def set_env(monkeypatch):
+    def _set_env(key, value):
+        monkeypatch.setenv(key, value)
+    return _set_env
+
+@pytest.fixture
+def mock_subprocess_run(mocker):
+    def _mock_subprocess_run(returncode=0, stdout="", stderr=""):
+        mock = mocker.patch('subprocess.run')
+        mock_return = Mock(spec=subprocess.CompletedProcess)
+        mock_return.returncode = returncode
+        mock_return.stdout = stdout
+        mock_return.stderr = stderr
+        mock.return_value = mock_return
+        return mock
+    return _mock_subprocess_run
+
+@pytest.fixture
+def create_dummy_config(temp_dir):
+    def _create_config(filename, content):
+        path = temp_dir / filename
+        path.write_text(content)
+        return path
+    return _create_config


Perry2004:
Hi SlncTrZ, thanks for the contribution! 

This would be good coverage that has been in our backlog for quite a while, yet we have not had time to complete it. 

However looks like the current PR status only contains test fixtures without actual cases. Is it still WIP? If you'd like to create a test suite, I can help incorporate that into the CI workflow.

Perry2004:
I'll close this since there's no further interaction in more than a week. Feel free to re-open if you have further updates.

PR #238, opened 2026-07-01
fix(security): reject path-traversal in extra_info paths

reacher-z:
## Reject path-traversal in `extra_info` paths

A task's `extra_info[].path` is untrusted input, but `copy_extra_info()` joined it to the case dir with **no containment check** (`src = task_dir / rel_path`). An absolute path or one containing `..` could copy **arbitrary host files** into the staged `my-info/` dir — which is then shipped inside a shareable **Harbor export** (`clawbench-harbor-adapt`) or bind-mounted into the agent container. This is a real information-disclosure vector.

### Fix
- New `validate_extra_info_path(task_dir, rel_path)` in `run_support/task.py`: rejects absolute paths and any target whose **resolved** location (symlinks followed) escapes `task_dir` (`src.resolve().relative_to(task_dir.resolve())`). Returns the unresolved `task_dir / rel_path`, so valid relative paths copy exactly as before.
- Applied in `copy_extra_info()` — benefits the native runner **and** the Harbor adapter, which both stage `extra_info`.

### Tests (all pass; ruff + pyright clean)
- valid relative path still resolves/copies;
- absolute path rejected;
- `..` escape rejected;
- `copy_extra_info` with `{"path": "../secret.txt"}` raises and copies nothing.

Small, self-contained, no behavior change for legitimate tasks. Found while building Harbor-runner tooling; extracted here as a standalone fix against `main`.

### DIFF
--- a/src/clawbench/runner/run_support/task.py
+++ b/src/clawbench/runner/run_support/task.py
@@ -129,6 +129,30 @@ def normalize_extra_info(raw: Any) -> tuple[list[dict[str, str]], list[str]]:
     return entries, warnings
 
 
+def validate_extra_info_path(task_dir: Path, rel_path: str) -> Path:
+    """Resolve an extra_info path and reject anything outside ``task_dir``.
+
+    A task's ``extra_info[].path`` is untrusted input. Without a guard, an
+    absolute path or one containing ``..`` could make ``copy_extra_info`` read
+    arbitrary host files into the staged my-info dir (which may be shipped in a
+    shareable export). Reject absolute paths and any target whose resolved
+    location — symlinks followed — escapes ``task_dir``. Returns the unresolved
+    ``task_dir / rel_path`` so normal copy semantics are unchanged.
+    """
+    if Path(rel_path).is_absolute():
+        raise ValueError(
+            f"extra_info path must be relative, got absolute: {rel_path!r}"
+        )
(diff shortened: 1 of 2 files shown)


Perry2004 (changes requested):
The new test case `test_validate_extra_info_path_rejects_absolute` is failing on Windows: https://github.com/TIGER-AI-Lab/ClawBench/actions/runs/28527654260/job/84568382985?pr=238

reacher-z:
@Perry2004 Fixed — thanks for catching this. Root cause: on Windows `Path("/etc/passwd").is_absolute()` is `False` (Windows needs a drive letter), so `validate_extra_info_path` rejected it via the containment check (`"escapes"`) rather than the `is_absolute` branch (`"absolute"`) — the path is still correctly rejected, but the test's `match="absolute"` didn't match the Windows rejection message.

Fixed in c5d8061 by relaxing the assertion to `match="absolute|escapes"` (both are valid rejections; the security property holds on all platforms). Also brought the branch up to date with `main`.

CI is now green on all OS including **pytest (windows-latest)**: https://github.com/TIGER-AI-Lab/ClawBench/actions/runs/28944758766

Ready for re-review when you have a moment.

Perry2004 (approved):
LGTM🫡

PR #246, opened 2026-07-03
docs(models): ship default judge in models.example.yaml (fixes #240)

### LINKED ISSUE #240: Default judge `deepseek-v4-pro` is unresolvable out-of-the-box
## Default judge `deepseek-v4-pro` is unresolvable out-of-the-box

`clawbench-run` and `clawbench-batch` default to `--judge deepseek-v4-pro` (`src/clawbench/runner/run.py:142`, and the batch runner). But `deepseek-v4-pro` is **not in `models/models.example.yaml`**, so a fresh clone that fills in one model key and runs with defaults hits `ERROR: model 'deepseek-v4-pro' not found` (or silently requires a paid DeepSeek key that's frequently out of credit → 402).

**Impact:** the documented "just run it" path fails on the scoring stage for new users.

**Fix options:**
1. Ship a `deepseek-v4-pro` stub entry in `models.example.yaml` (commented), and/or
2. Change the default to `--no-judge` (Stage-1 only) so out-of-box runs succeed, judging opt-in, or
3. Fail with an actionable message pointing at the judge setup when the judge model is absent.

reacher-z:
## Ship the default judge in `models.example.yaml` (fixes #240)

`clawbench-run` / `clawbench-batch` default to `--judge deepseek-v4-pro` (`run.py:142`), but `models/models.example.yaml` never contained it. A fresh clone that copies the example, fills in one agent key, and runs with defaults fails at the **scoring stage** with `ERROR: model 'deepseek-v4-pro' not found`.

**Fix:** add a commented `deepseek-v4-pro` entry to the example (with `base_url`/`api_type`), plus a note that you can point `--judge` at another entry or pass `--no-judge` for interception-only scoring. No code change; unblocks the out-of-box path.

Closes #240.

### DIFF
--- a/models/models.example.yaml
+++ b/models/models.example.yaml
@@ -14,7 +14,7 @@ qwen3.5-397b-a17b:
 # openai-completions / openai-responses / anthropic-messages api_type.
 deepseek-v4-pro:
   api_key: "sk-..."
-  base_url: https://api.deepseek.com
+  base_url: <api_base_url>
   api_type: openai-completions
 
 # For multiple API keys (round-robin), use api_keys instead of api_key:
--- a/models/models.example.yaml
+++ b/models/models.example.yaml
@@ -7,6 +7,16 @@ qwen3.5-397b-a17b:
   api_type: openai-completions
   thinking_level: medium             # optional
 
+# LLM JUDGE (Stage-2 scoring). clawbench-run/clawbench-batch default to
+# `--judge deepseek-v4-pro`, so it must exist here or the scoring stage fails
+# with "model not found". Fill in a key below, point --judge at another entry,
+# or pass --no-judge to skip Stage-2 (interception-only). The judge must be an
+# openai-completions / openai-responses / anthropic-messages api_type.
+deepseek-v4-pro:
+  api_key: "sk-..."
+  base_url: https://api.deepseek.com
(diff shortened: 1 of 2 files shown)


models/models.example.yaml
```diff
@@ -7,6 +7,16 @@ qwen3.5-397b-a17b:
...
   thinking_level: medium             # optional
 
+# LLM JUDGE (Stage-2 scoring). clawbench-run/clawbench-batch default to
+# `--judge deepseek-v4-pro`, so it must exist here or the scoring stage fails
+# with "model not found". Fill in a key below, point --judge at another entry,
+# or pass --no-judge to skip Stage-2 (interception-only). The judge must be an
+# openai-completions / openai-responses / anthropic-messages api_type.
+deepseek-v4-pro:
+  api_key: "sk-..."
+  base_url: https://api.deepseek.com
```
Perry2004: Hardcoding the `base_url` requires the user to have an official api.deepseek.com key.
Leave the `base_url` empty so users can use APIs other than the official ones. 
```suggestion
  base_url: <api_base_url>
```

Perry2004 (changes requested):

Perry2004 (approved):

Issue #159, opened 2026-05-13
feat(analysis): error analysis + per-task failure breakdown + agent-capability insights from results


reacher-z:
## Motivation

The leaderboard answers \"which model wins\" but it's silent on **why** — why does GLM-5.1 at 48.5% Intercepted on V2 still only convert 18.5% into Reward? Why did Claude Sonnet 4.6 fail tasks #043 (Rover) and #097 (Eventbrite) but pass the visually-harder #001 (Uber Eats)? Which **failure modes** are the leaderboard's 0% reward agents hitting first? With 2,326 graded trajectories ([5-layer bundles](https://huggingface.co/datasets/TIGER-Lab/ClawBenchV2Trace)) we have the data to answer this — we just haven't surfaced it.

This issue is to ship a structured analysis layer on top of the existing run data, both for the **paper rebuttal / discussion section** and for the public website.

### A. Error taxonomy & failure-mode classification

Build a labeled corpus of **why a run failed**. For each failed run (or partial run), classify into:

1. **Navigation failure** — agent couldn't find the right page (wrong site, infinite menu loops, blocked by paywall/login)
2. **Form-fill error** — reached the form, submitted wrong content (typo'd field, missing required field, used wrong persona info)
3. **Selector / DOM mismatch** — agent's `actions.jsonl` clicked the wrong element (e.g., \"Search\" button vs \"Filter\" button)
4. **Auth / captcha block** — agent couldn't get past login, MFA, or human-verification
5. **Schema mismatch** — request body shape doesn't match the rubric (e.g., `delivery_address` vs `shipping_address`)
6. **Time-out / queue cap** — agent didn't reach the final action within the per-task time budget
7. **Hallucinated success** — agent claimed task done without ever submitting (Stage 1 false, agent thinks Stage 1 true)
8. **Reasoning collapse** — agent gives up / loops / produces \"I cannot complete this task\" without exhausting the action budget

Method: use the existing `deepseek-v4-pro` judge (or a fresh one) but re-prompted to **label the failure category** given the trajectory + the rubric's expected payload. Cross-check against a hand-labeled sample (50 runs).

Output: a new derived dataset `TIGER-Lab/ClawBench-Errors` with one row per failed (run, task) — model, harness, task_id, failure_category, judge_reason, evidence_step_id.

### B. Per-task failure attribution (\"why didn't model X solve task Y\")

For every (model, task) failure, surface a structured **diagnosis card**:

```
[...]

PR #256, opened 2026-07-10
feat(harness): null baseline harness — the benchmark floor (closes #245)

### LINKED ISSUE #245: Publish null/random baseline + interceptor false-positive rate as validity evidence
## Publish a null/random baseline + interception false-positive rate as validity evidence

A live-web, interception-scored benchmark needs to show it isn't luck-passable and that the interceptor doesn't false-fire. We have the data to publish this cheaply:

- **Null / do-nothing baseline:** across 820 real runs, 129 took 0 actions; the interceptor fired on **0** of them → a do-nothing agent scores ~0.
- **Interceptor false-positive:** no 0-action run was ever marked intercepted → Stage-1 does not trigger without real agent activity.

**Ask:**
1. Add an explicit **null agent** (does nothing) and a **random-click agent** as leaderboard baselines — both should score ~0, establishing the floor.
2. Report the interceptor's false-positive rate (interceptions on runs with no genuine task-completing action).
3. Surface these in the methodology section as validity evidence (this is exactly the kind of check reviewers ask for and competitors like DeepSWE publish).

reacher-z:
## Null baseline harness — the benchmark floor (closes #245)

Adds a `null` harness: a do-nothing agent that connects to nothing and takes no browser action, so the run records **0 actions/requests and the interceptor never fires**. A null agent must therefore score **~0**.

**Why it matters (the methodological artifact reviewers ask for):**
- Establishes the **benchmark floor** — proves ClawBench isn't luck-passable.
- Directly demonstrates the **interceptor false-positive rate** — Stage-1 must not fire without genuine agent activity. (The rate is computed by the new `clawbench-analyze` tool, #159/#255.)

**What's here**
- `runtime/harnesses/null/{Dockerfile.null, setup-null.sh, run-null.sh, usage-emitter.py}` — `FROM clawbench-base`, makes no model call (no API key needed).
- Registered in `harnesses.yaml` (name quoted to dodge the YAML `null` keyword) + `test_harness_registry` expectations.

**Verified:** registry test passes (`null` registered); `clawbench-null` **builds** from `clawbench-base`; composition check confirms the scripts are present and `run-null` is a genuine no-op. Full suite green; ruff clean.

Run it: `clawbench-run --harness null --task <id>` → expect 0 actions, `intercepted: false`, reward 0. A follow-up can add a `random-click` baseline for a non-trivial floor.

### DIFF
--- a/src/clawbench/runtime/harnesses/harnesses.yaml
+++ b/src/clawbench/runtime/harnesses/harnesses.yaml
@@ -112,3 +112,14 @@ harnesses:
         path: /data/agent-messages.jsonl
       - type: file
         path: /data/agent-messages.raw.jsonl
+  # Null baseline — takes no action; establishes the benchmark floor (~0 score,
+  # and the interceptor must not fire without genuine agent activity).
+  - name: "null"
+    image: clawbench-null
+    dockerfile: "null/Dockerfile.null"
+    setup_script: "null/setup-null.sh"
+    run_script: "null/run-null.sh"
+    usage_emitter: "null/usage-emitter.py"
+    agent_message_sources:
+      - type: file
+        path: /data/agent-messages.jsonl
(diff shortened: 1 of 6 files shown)


reacher-z:
Superseded by the combined null + random-click baselines PR (folds this null harness in + adds random-click). Closing in favor of that.

Issue #244, opened 2026-07-02
Local v2 corpus is 129 tasks vs published 130 — reconcile and pin the count


reacher-z:
## Local v2 corpus is 129 tasks but published count is 130 — reconcile + pin

`test-cases/v2/` contains **129** task dirs, but the published v2 size is **130** (README/HF card/site). One task is missing from the checked-out corpus, and the advertised count is also inconsistent across surfaces (README "130", HF card, site "163 platforms", etc.).

**Impact:** anyone reproducing "v2" runs 129, not the canonical 130 — silent ~1-task divergence in every reported denominator; and the headline counts don't agree with each other or with the data.

**Ask:**
1. Restore the missing 130th v2 task (or correct the published count to 129 if 130 was aspirational).
2. Add a CI check asserting `len(test-cases/v2) == <canonical>` so the corpus size can't silently drift.
3. Reconcile the task/site count into one canonical triple stated identically in README, HF card, and site. (Related to #216.)

Perry2004:
It should be pinned to 129. That one additional task was removed after the initial release due to issues.

Perry2004:
Task changes on released suites are unusual, thus I believe a dedicated test case/GHA is not needed.

I'll keep in mind to update the counts if any changes are made on existing corpuses in the future.

Perry2004:
Readme counts have already been updated.

PR #258, opened 2026-07-11
test(corpus): guard task-target integrity (non-empty valid-regex url_pattern + no dupes)

reacher-z:
Adds corpus-integrity CI guards (re #244).

`validate_task_data` checks `eval_schema.url_pattern` is a *string* but not that it's **non-empty** or a **valid regex** — so two silent-corruption classes slip through:
- **empty** `url_pattern` → `re.search('', url)` matches **every** request → false-positive Stage-1 interceptions (task always 'passes' Stage-1)
- **malformed regex** → the runtime interceptor crashes at eval time

New tests over every suite (v1/v2/v1-lite/claw-eval): `url_pattern` is non-empty + compiles as a regex, and no two tasks share the same `(url_pattern, method, instruction)` (accidental copy-paste). Both **pass on the current corpus** and act as a drift/corruption guard.

(The 129-vs-130 count reconciliation in #244 needs the source data — left to maintainers.)

### DIFF
--- a/tests/test_host_tasks.py
+++ b/tests/test_host_tasks.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import re
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


Perry2004 (changes requested):
A `validate-task` GHA is already implemented for similar purposes. If additional restrictions need to be implemented, please add to that GHA instead. 

Since this is not related with code logic and thus should be in the CI rather than the general test file.

PR #273, opened 2026-07-31
feat(harness): add WebBrain extension agent

### LINKED ISSUE #272: feat(harness): add a WebBrain browser-extension agent
## Summary

Would you be open to a dedicated `webbrain` harness for evaluating
[WebBrain](https://github.com/webbrain-one/webbrain), a browser-native Chrome extension agent,
on ClawBench tasks?

This would add a meaningfully different agent architecture to the current CLI/framework and
Claude-in-Chrome harness set while reusing ClawBench's existing browser recorder and evaluation
pipeline.

## Why this fits the current architecture

I did a static source pass over both projects:

- ClawBench's harness registry is data-driven in
  `src/clawbench/runtime/harnesses/harnesses.yaml`, and the base entrypoint already accepts
  additional unpacked extensions through `EXTRA_LOAD_EXTENSIONS`.
- WebBrain's `test/anonymous/run.mjs` demonstrates the external control path: configure a
  provider in `chrome.storage.local`, find the task tab, then send
  `{ target: "background", action: "chat", tabId, text, mode: "act" }` from an extension page.
  `src/chrome/src/background.js` owns that `chat` handler.
- I searched open/closed issues and PRs in this repository for `WebBrain` and found no existing
  proposal or implementation.

No WebBrain source needs to be copied into ClawBench. ClawBench is Apache-2.0 and WebBrain is
MIT, so the preferred boundary would be a pinned upstream release/commit loaded as an unpacked
extension.

## Proposed minimal vertical slice

1. Add a `webbrain` entry to the harness registry and a small harness image pinned to a WebBrain
   release or commit.
2. Load WebBrain alongside ClawBench's recorder with `EXTRA_LOAD_EXTENSIONS`.
3. Initially support ClawBench's OpenAI-compatible API configuration (`MODEL_NAME`, `BASE_URL`,
   `API_KEY`) by writing one configured WebBrain provider to extension storage. Fail fast with a
   clear message for unsupported API types.
4. Use CDP to locate WebBrain's MV3 service worker/extension ID and the active task tab, then run
   the instruction through the existing `background/chat` interface in `act` mode.
5. Emit the returned result/updates and terminal errors to `/data/agent-messages.jsonl`, plus a
   usage row when WebBrain exposes sufficient token data.
6. Add deterministic tests for provider mapping, CDP target selection, timeout/error handling,
   JSONL output, and registry validation; add one opt-in container smoke test on a non-consequential
   task.

## Design questions

Before implementing, I would like to align on four boundaries:

1. **API formats:** is an OpenAI-compatible-only v1 acceptable, or should every new harness
   support all ClawBench API types from day one? My recommendation is the smaller OpenAI-compatible
   slice, with explicit rejection for the others.
2. **Versioning:** should the image pin a WebBrain release tag or an exact commit? I lean toward an
   exact commit plus checksum so benchmark runs remain reproducible.
3. **Consequential actions:** WebBrain deliberately asks for confirmation before some submissions.
   I do not propose a blanket bypass. The first smoke test can be read-only/non-consequential;
   confirmation bridging can be designed separately if the corpus requires it.
4. **Usage accounting:** if the returned trace lacks complete token counts, is a clearly marked
   unavailable usage record acceptable for v1, or is exact usage accounting a merge requirement?

## Acceptance criteria

- `clawbench run ... --harness webbrain` launches without changing existing harness behavior.
- A safe task reaches WebBrain through its public extension message path and produces a valid
  ClawBench agent transcript.
- Missing extension targets, unsupported API types, model errors, and timeouts fail deterministically
  and leave diagnostic output.
- Core tests run without external model credentials; the real container/model test remains opt-in.

If this scope and ownership look right, I can implement the vertical slice and tests.

alectimison-maker:
## What does this PR do?

Adds a `webbrain` harness that evaluates the pinned
[WebBrain](https://github.com/webbrain-one/webbrain) MV3 browser extension on ClawBench tasks.

The harness:

- loads an exact WebBrain commit alongside the existing ClawBench recorder;
- maps an `openai-completions` model config into WebBrain's persisted provider contract;
- invokes WebBrain's existing `background/chat` extension interface through CDP;
- writes WebBrain updates, structured trace events, the final response, and normalized token
  usage into ClawBench's standard data artifacts; and
- participates in the existing registry, image build, watchdog, interceptor, and recording
  lifecycle without changing other harnesses.

## Motivation and design

ClawBench currently covers CLI, framework, and bridged extension agents, but it cannot evaluate a
self-contained browser-extension agent through the same corpus and five-layer recording pipeline.
The existing `EXTRA_LOAD_EXTENSIONS` seam and data-driven harness registry make a dedicated adapter
smaller than introducing a new generic protocol.

WebBrain is pinned to commit `7b2357502da03d931117404bb89a60006bc3fa97`; its source is loaded
unpacked and no WebBrain repository scripts or dependencies run during the image build. The driver
uses the base runtime's existing `websocket-client` dependency, so the harness adds no Python
package dependency.

Unattended runs use WebBrain's supported Instant clarification setting rather than patching its
safety logic. The benchmark instruction is the user request, while ClawBench's interceptor remains
the final boundary for consequential network submissions.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

## Test plan

- [x] `uv run --frozen pytest tests -q` — 166 passed
- [x] `uv run --frozen ruff check .` — passed
- [x] `uv run --frozen ruff format --check .` — passed
- [x] `uv run --frozen pyright src/clawbench tests` — passed
- [x] `bash -n src/clawbench/runtime/harnesses/webbrain/{setup-webbrain.sh,run-webbrain.sh}` — passed
- [x] Python byte-compilation for the driver and usage emitter — passed
- [x] `uv build` plus `twine check` and wheel-content inspection — passed
- [ ] Local Docker/Podman image build — unavailable because neither container engine is installed
  in the development environment; the changed-harness CI job exercises the real Docker build.

## Compatibility and risks

- v1 supports local browser mode and `api_type: openai-completions`; unsupported modes/types fail
  before a model request.
- `api_keys` currently selects its first entry rather than rotating keys.
- Model vision support is not present in ClawBench's model schema, so the generic WebBrain provider
  stays text/tool-only to avoid sending image blocks to incompatible endpoints.
- Cost is reported as `price_unavailable`; token counts are preserved from WebBrain trace usage.
- A real model-backed browser smoke run was not performed locally because it requires a container
  engine and model credentials.

## Related issues

Closes #272


Perry2004:
Thanks for the PR. I'll look into this. 🙌

Perry2004 (changes requested):
Hi @alectimison-maker, thanks for the PR!

I reviewed the new integration and found a couple of issues:

1. The agent seems to be stuck on the initial `about:blank` page and is unable to perform any actions.
2. The stop reason is reported incorrectly. A run in which no request is intercepted is classified as `"stop_reason": "eval_matched"`.

```json
{
  "intercepted": false,
  "stop_reason": "eval_matched",
  "stop_description": "Session stopped: eval_matched.",
  "request": null,
  "schema": {
    "url_pattern": "https://never-match.invalid/",
    "method": "GET"
  }
}
```

To reproduce:

-  Create a dummy task:

```json
{
  "instruction": "Go to Google Search and search for clawench. Do not click any search results.",
  "eval_schema": {
    "url_pattern": "https://never-match.invalid/",
    "method": "GET"
  },
  "time_limit": 2
}
```

-  Run the new WebBrain harness.

I suspect the stalling may be caused by a permission/consent/mode issue, as the extension's settings page opens when the agent starts and when I manually install the extension on a new profile it needs me to conscent a few things and select the mode:

```json
{
  "type": "pageLoad",
  "timestamp": 1785794425648,
  "url": "chrome-extension://opnkcadpbfggglmiofmahfblhlghklob/src/ui/settings.html",
  "title": "WebBrain Settings"
}
```

I'm not familiar with this agent, so my suspicions may be incorrect. Please let me know if you're able to reproduce and resolve these issues.

alectimison-maker:
Thanks for the detailed reproduction. I was able to trace and address both issues in f8e6d8f:

- unattended runs now preconfigure WebBrain's first-run onboarding and prompt-free permission mode, in addition to the existing Instant clarification setting, so the settings/consent flow no longer leaves the agent on `about:blank`;
- `.stop-requested` is now classified as `eval_matched` only when `interception.json` explicitly contains `"intercepted": true`; missing, invalid, or `false` interception data reports `stop_requested` instead.

I added regressions for the storage configuration and all three stop-classification cases (missing/false/true, plus invalid JSON). Revalidated the current head with:

- `uv run --frozen pytest tests/test_webbrain_harness.py -q` - 12 passed
- targeted Ruff check and format check - passed
- `git diff --check` - passed

A live container/model-backed run still requires the repository CI environment because this local host has no Docker/Podman engine or model credentials.

Perry2004:
Thanks for the changes! I'll re-review it in a bit.

src/clawbench/runtime/harnesses/webbrain/run-webbrain.sh
```diff
@@ -0,0 +1,93 @@
...
+if [ -z "$STOP_REASON" ]; then
+  if kill -0 "$AGENT_PID" 2>/dev/null; then
+    echo "Time limit (${MAX_WAIT}s) exceeded, stopping WebBrain."
+    STOP_REASON="time_limit_exceeded"
+  else
+    wait "$AGENT_PID" || AGENT_STATUS=$?
+    if [ "$AGENT_STATUS" -eq 0 ]; then
+      STOP_REASON="agent_exited"
+    else
+      STOP_REASON="agent_error"
```
Perry2004: The new `agent_error` reason is not included in https://github.com/TIGER-AI-Lab/ClawBench/blob/f6cbb854b159065a3443d1fdaf4b3cb6356fad93/src/clawbench/runner/run_support/results.py#L12-L23. 
It'd be great to change it to `webbrain_failed` and record in `results.py`.

Perry2004 (changes requested):
Thanks for the changes, this time it is very close to a complete integration, other than a few small issues. 
1. The agent error stop reason is not recorded in `results.py`. 
2. When the run is intercepted by the interception, the `agent-messages` only contain the initial user input. 
3. In `run-meta.json`, the `run_metrics.usage.status` shows `usage_unavailable` with all 0 values. 

I think both issues are somewhere in `run-webbrain-agent.py` since that's the transcript & usage recording happens though I might be wrong. 
For non-intercepted runs, I believe the `agent-messages` are recorded correctly, so might be the SIGTERM handling issue. 
To reproduce, you may use this dummy task: 
``` json
{
  "instruction": "Open https://httpbin.org/forms/post. Fill the customer name with Clawbench User, set telephone to 1234567, set email to clawbench@example.com, choose Medium pizza size, check the Bacon topping, and submit the order. Stop after submitting.",
  "eval_schema": {
    "url_pattern": "https://httpbin\\.org/post(?:[?#].*)?$",
    "method": "POST"
  },
  "time_limit": 4
}
```

Perry2004 (approved):
The PR looks good to me. 
I'll merge it and include in the next release. 🫡 
Thanks for the contribution! 🎉

alectimison-maker:
Thanks！

Issue #130, opened 2026-05-08
RELEASE: refresh website / code / HF / arXiv v2 with v2 corpus, new models, Hermes, and error analysis


reacher-z:
## Goal

Coordinated refresh across all four public surfaces so they reflect the current state of ClawBench (post-v2-corpus, post-Hermes, post-#129 eval sweep).

## Surfaces

- [ ] **Website** — landing page + leaderboard
- [ ] **Code** (this repo) — README, docs, examples
- [ ] **HuggingFace** — [dataset page](https://huggingface.co/datasets/NAIL-Group/ClawBench) + [paper page](https://huggingface.co/papers/2604.08523)
- [ ] **arXiv v2** — replacement submission of [2604.08523](https://arxiv.org/abs/2604.08523)

### 1. New models
The 8-model lineup from #129:
- GPT-5.5, Claude Opus 4.7, Gemini 3.1, M2.7, Kimi K2.6, GLM 5.1, DeepSeek V4 Flash, DeepSeek V4 Pro

### 2. New data
- **V2 corpus** (130 tasks, `test-cases/v2/`) — landed via #116
- Updated category breakdown and site coverage stats

### 3. New Hermes agent harness
- HermesAgent runner (alongside OpenClaw and Claude Code)
- Setup + run instructions in README and on the website

### 4. New analysis
- Cross-harness comparison (OpenClaw / HermesAgent / Claude Code) on the same model
- v1 ↔ v2 deltas per model
- Cost / wall-clock vs. accuracy frontier

### 5. Error analysis
- Failure taxonomy (auth walls, captcha, navigation loss, hallucinated actions, judge disagreement, etc.)
- Per-category and per-site failure rates
- Representative failure traces (anonymized) with commentary

### Website
- [ ] Refresh leaderboard with #129 results (3 harnesses × 2 corpora × 8 models)
- [ ] Add v2 corpus section + category/site stats
- [ ] Add Hermes harness card
- [ ] New "Analysis" page with the 4 cross-cuts above
- [ ] New "Error analysis" page with taxonomy + traces

### Code (repo)
- [ ] README leaderboard table refreshed
- [ ] Hermes setup/run docs polished (`run-hermes.sh`, `setup-hermes.sh`)
- [ ] `eval-results/` populated with the 48-cell sweep
- [ ] Analysis notebooks/scripts checked in under `eval/`
- [ ] News entry for the release date

### HuggingFace
[...]

PR #288, opened 2026-08-17
feat: add LiteLLM judge api_type

prodmanpd:
## What does this PR do?

Adds a `litellm` judge `api_type`, so the ClawBench judge can reach any of 100+ providers (OpenAI, Anthropic, Gemini, Bedrock, Vertex, Azure, Groq, ...) through a single call instead of a per-wire-format raw-HTTP function. `litellm.completion` natively speaks the OpenAI / Anthropic / Gemini / Bedrock protocols, so it collapses the existing per-`api_type` handlers into one path, and `drop_params=True` keeps a single call portable across providers.

Changes:
- `src/clawbench/runner/judge.py` and `judge_llm.py` - new `_call_litellm(...)` + a `litellm` branch in each judge's dispatch (mirrors the sibling `_call_openai_chat` / `_call_anthropic_messages`).
- `src/clawbench/runner/run_support/api_preflight.py` - `litellm` preflight validation (a tiny call before containers run).
- `src/clawbench/tui.py` - `litellm` added to `API_TYPES` + a LiteLLM provider preset.
- `pyproject.toml` - optional extra `litellm = ["litellm>=1.85.0,<2.0"]` (lazy-imported, so the base install is unaffected).
- `tests/test_judge_litellm.py` - dispatch / kwargs / preflight / import-error coverage.

Additive and opt-in: existing `api_type` handlers are untouched; `litellm` is an optional dependency imported lazily inside `_call_litellm`, with a clear ImportError pointing to `pip install clawbench[litellm]` if the api_type is used without it.

## Corpus

- [ ] v2
- [ ] v1
- [x] both
- [ ] not applicable

## Test plan

- [x] Verified against the repo's frozen `uv` environment:
  - `uv run --frozen pytest tests/test_judge_litellm.py` -> **6 passed** (dispatch in both judges, `drop_params=True`, credentials forwarded only when set, preflight, and the ImportError path).
  - `uv run --frozen ruff check .` -> `All checks passed!`; `uv run --frozen ruff format --check .` -> clean; `uv run --frozen pyright <changed files>` -> **0 errors**.
  - Live E2E: ran the real `judge_llm.judge_request` with `api_type: "litellm"` routed through a LiteLLM proxy (`gpt-4o-mini`). Aligned instruction -> `{"match": true, ...}`; contradicting instruction -> `{"match": false, "reason": "Instruction specifies 'BLUE shirt', request body shows 'item: red shirt'"}`. This exercises the full chain: `judge_request` -> `api_type` dispatch -> `_call_litellm` -> `litellm.completion` -> provider -> parsed verdict.
  - Pin resolves: `litellm>=1.85.0,<2.0` installs against the current latest stable (1.97.0).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,11 @@ dependencies = [
     "rich>=13.0",
 ]
 
+[project.optional-dependencies]
+# Optional: judge api_type "litellm" routes to 100+ providers via one call.
+# Lazy-imported, so the base install stays dependency-light.
+litellm = ["litellm>=1.85.0,<2.0"]
+
 [project.urls]
 Homepage = "https://claw-bench.com"
 Repository = "https://github.com/reacher-z/ClawBench"
--- a/src/clawbench/runner/judge.py
+++ b/src/clawbench/runner/judge.py
@@ -172,6 +172,40 @@ def _call_anthropic_messages(
     return resp["content"][0]["text"]
 
 
+def _call_litellm(model_cfg: dict, model_name: str, system: str, user: str) -> str:
+    """Route the judge through LiteLLM, reaching any of 100+ providers with one
+    call. ``model_name`` is a LiteLLM model string (e.g. ``gpt-4o-mini``,
+    ``anthropic/claude-sonnet-4-6``, ``gemini/gemini-2.5-flash``); set
+    ``base_url`` to point at a LiteLLM proxy."""
+    try:
+        import litellm  # type: ignore
+    except ImportError as e:
+        raise ImportError(
+            "The 'litellm' package is required for api_type 'litellm'. "
+            "Install it with 'pip install clawbench[litellm]' or 'pip install litellm'."
+        ) from e
+
+    kwargs: dict[str, Any] = {}
+    if model_cfg.get("base_url"):
+        kwargs["api_base"] = model_cfg["base_url"]
+    if model_cfg.get("api_key"):
+        kwargs["api_key"] = model_cfg["api_key"]
+    resp = litellm.completion(
+        model=model_name,
+        messages=[
+            {"role": "system", "content": system},
+            {"role": "user", "content": user},
+        ],
+        max_tokens=4096,
+        temperature=0,
+        # Silently drop kwargs a given provider doesn't accept so one call works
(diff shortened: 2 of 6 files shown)


Perry2004:
Thanks for the contribution. 

Could you clarify the use case for adding a LiteLLM API type? ClawBench already have support for OpenAI, Anthropic, and Google GenAI API types and custom URLs, which should cover almost all cases calling LLM APIs I believe. 

Can you explain the motivation why a LiteLLM is needed for calling LLM APIs for the judge.

Perry2004:
Closing the PR as no activities after half a month. Feel free to re-open if needed.

liuzhao1225:
I'm concerned about the supply-chain boundary of this LiteLLM integration. The [upstream incident report](https://docs.litellm.ai/blog/security-update-march-2026) confirms compromised PyPI releases 1.82.7/1.82.8 in March 2026; those releases were removed.

At reviewed head `a5d00b9d2b`, this PR includes `litellm>=1.85.0,<2.0` in `pyproject.toml`.

That upper bound permits later 1.x releases beyond a single tested version when dependencies are resolved afresh or upgraded. Please document tested versions, artifact verification, and whether installation of the SDK is opt-in. A disabled application feature cannot prevent an installed malicious Python-startup hook.

This is one of [49 observed LiteLLM integration PRs](https://github.com/liuzhao1225/litellm-integration-supply-chain-review/blob/main/PR-INDEX.md) from the same account, whose [author acknowledged the cross-project effort](https://github.com/chartbrew/chartbrew/pull/365#issuecomment-5207856315). The pattern prompted this review; it does not establish malicious intent or connect the author to the incident. The consolidated questions and corrections are in [YouDub #130](https://github.com/liuzhao1225/YouDub-webui/pull/130); this PR's [review snapshot](https://github.com/liuzhao1225/litellm-integration-supply-chain-review/blob/main/prs/TIGER-AI-Lab__ClawBench__288.json) preserves the revision and scope.

This is a follow-up documentation note on the reviewed historical change; it does not assert that the current release is compromised.

Perry2004:
Hi @liuzhao1225, thanks for the heads-up. The PR has already been closed since it is not needed and I'll keep an eye on such patterns.

PR #282, opened 2026-08-15
README declutter: promote Awesome Works, weekly citation scan, fold long sections

reacher-z:
Implements the maintainer's README review (decision log D1–D9, exported 2026-08-15).

## What changed

**Awesome Works (D2, D3)**
- Section moved from line ~981 (after both FAQ blocks) to directly after **News**, where readers actually see it; institution-emphasis line bolded at the top.
- **5 new citing papers added** (all arXiv-verified): From Chatbot to Digital Colleague (Tencent Youtu Lab, Tsinghua & SYSU), PAST-Bench (Princeton & UChicago), SecRespond (Tongyi Lab, Alibaba Cloud & HKUST), EnterpriseClawBench (Horizon Research & Frontis.AI), ContinualSkillBench (PKU & BIGAI).
- ⚠️ Reviewer notes: (1) **D2 selections were given but the overall verdict was left unset** in the review export — please double-check the 5 additions; 3 candidates were excluded per the maintainer's editorial criteria. (2) PAST-Bench's PDF prints no affiliations; Princeton/UChicago is inferred from the senior authors (Gen-Verse org) — please confirm before merge.

**Auto-update pipeline (D1)**
- `scripts/update_awesome_works.py` + `.github/workflows/awesome-works.yml`: every Sunday, fetch citing papers of arXiv:2604.08523 from Semantic Scholar, diff against README + previously filed scan issues, and open an `awesome-works` issue cc @Perry2004 for curation. Never edits the README directly. Dry-run already found 10 candidates (incl. ClawTrack, MyPCBench, Harness-Bench), so expect a first issue on the first scheduled run.

**Declutter (D4–D8)**
- Logo: side-by-side hero (image left 200px, tagline right); source PNG compressed 2.0 MB → 140 KB.
- News: 7 newest visible, 7 older under an *Earlier news* fold.
- Badges: 9 core visible (Star / arXiv / Daily Paper / Dataset / Trace / claw-bench.com / PyPI / **Discord** / License — Discord & Trace kept per review); counters, Codespaces, DeepWiki moved into the existing fold.
- CLI: 3 common commands visible, the rest under *All commands*; Reproduce Path A/B and the Harbor quick-start block folded (scope extension of D6 in the same spirit — flag if unwanted).
- Duplicates removed: the two FAQ sections merged into one details-style FAQ (`#frequently-asked-questions` anchor preserved; no unique Q&A dropped), HarnessBench banner and the second clone one-liner removed.

**zh-CN (D9)**
[...]

PR #289, opened 2026-08-18
docs: surface Awesome Works above the fold (it sits at 85% depth, below the FAQ)

reacher-z:
Answers a maintainer question — *does moving Awesome Works up help publicity?* — with a measurement, then acts on it.

## The measurement (on current `main`)

Awesome Works sits at **line 981 of 1,151 — 85% of the way down, below both FAQ blocks**. A reader reaches it only after ~40 screens of scrolling, and the FAQ ahead of it is a classic exit point: people who open a FAQ are usually after one answer and then leave.

That is the wrong place for the strongest piece of social proof the project owns — 17 works from Google DeepMind, Stanford, UC Berkeley, Microsoft Research, Harvard, ETH Zürich, Oxford, Northwestern, ByteDance Seed and HKUST, plus an industry model report (Li Auto's Mach-Mind-4-Flash) that evaluates on ClawBench. For a benchmark, *who already builds on this* is the credibility argument that decides whether a new reader adopts it. Buried below the FAQ, most readers never see it.

## The change

- **Awesome Works → line 199, right after News.** It now sits in the same reading zone as Datasets and Results, where a reader deciding whether to adopt ClawBench is still on the page.
- **The institution line becomes the lede**, in bold, and names the Li Auto report — one sentence a skimmer absorbs without reading a single entry.
- **8 entries visible, 9 behind a `<details>` fold** — the section costs little vertical space up top. All 17 entries preserved; the `#awesome-works-using-clawbench` anchor is unchanged, so existing links still work.
- **License badge** moves into the existing curated-lists fold, leaving only signal badges in the visible row (maintainer request).
- **zh-CN contact** now reads "Yuxuan Zhang" without the CJK name (maintainer request).

## Notes for review

- Trade-off worth your opinion: this pushes the feature strip and Datasets ~45 lines down. I judged first-read credibility to outweigh that, but flipping the order is a one-line change if you disagree.
- **Overlap:** #282 also moves this section (approved as D3 in the maintainer's earlier review) and grows it to 26 entries; #283 rewrites the README around it but leaves it at 80% depth. Whichever lands first, I'll rebase the others onto it — no content will be lost. If you prefer, merge this one first since it is the smallest single-purpose diff of the three.
- `<details>` balanced (19/19); anchors verified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,6 @@
 [![PyPI version](https://img.shields.io/pypi/v/clawbench-eval?style=flat-square&logo=pypi&color=3775A9&logoColor=white)](https://pypi.org/project/clawbench-eval/)
 [![Ask a question](https://img.shields.io/badge/Ask%20a%20question-181717?style=flat-square&logo=github&logoColor=white)](https://github.com/TIGER-AI-Lab/ClawBench/issues/new/choose)
 [![GitHub stars](https://img.shields.io/github/stars/TIGER-AI-Lab/ClawBench?style=flat-square&logo=github&color=181717&cacheSeconds=300)](https://github.com/TIGER-AI-Lab/ClawBench)
-[![License](https://img.shields.io/github/license/TIGER-AI-Lab/ClawBench?style=flat-square&color=A42E2B)](https://github.com/TIGER-AI-Lab/ClawBench/blob/main/LICENSE)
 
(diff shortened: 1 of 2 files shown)


reacher-z:
Rebased onto `main` now that #283 has landed — the measurement holds on the new README: Awesome Works sits at **line 823 of 1,033 (80% down), immediately after the FAQ**. This PR moves it to line 438, right after News. All 17 entries and the anchor are preserved; 8 visible + 9 folded. `<details>` balanced 29/29.

PR #291, opened 2026-08-18
ci: fail static-check on broken markdown links (fixes a broken README image + 2 dead script refs)

reacher-z:
Adds the quality gate that would have caught three link breaks currently live on `main` — and fixes those three.

## What it found

| Break | Effect |
|---|---|
| `README.md` — "Reproduce the leaderboard" heading uses `assets/icons/check-double.svg` | The file does not exist; the heading renders a **broken image** on the repo front page |
| `docs/scoring.md` — "run `scripts/clawbench_rescore.py`" (link + code block + reference list) | That path has never existed. Readers following our reproducibility doc hit "No such file" |
| `eval/scoring.md` — same three references | Same |

The rescore entry point is the `clawbench-rescore` console script (`src/clawbench/eval/rescore.py`), so both docs now say that and link the module. The icon reference is repointed at the existing `shield-halved.svg`.

## The gate

`scripts/check_markdown_links.py` walks every tracked `*.md`, resolves repo-relative `[text](path)`, `alt` and `` targets against the filesystem, and exits 1 listing `file:line: target` for each break. External URLs (http/https/mailto/data/tel) and pure anchors are skipped — this guards against renames and deletions, not against the network, so it stays fast and never flakes. Wired as the first step of `static-check`.

Verified both ways: clean tree → `all 128 repo-relative markdown links resolve` (exit 0); with a deliberately planted bad link → exit 1 naming the file and line.

## Note on overlap

#283 independently spotted the missing icon and fixes it by **adding** `check-double.svg`. If #283 lands first, drop my one-line icon change during rebase and keep theirs — the checker and the two `scripts/clawbench_rescore.py` fixes are unaffected either way.


reacher-z:
Rebased onto `main` now that #283 has landed. As flagged in the description, #283 fixed the broken README icon by adding `assets/icons/check-double.svg`, so I dropped my one-line icon change and kept theirs — this PR is now just the checker plus the two `scripts/clawbench_rescore.py` doc fixes. Post-rebase: **all 154 repo-relative links resolve**.

Perry2004 (changes requested):
The markdown check shouldn't be added to the existing `static-check` GHA since that won't even be triggered by changes in the docs/markdown.

Perry2004 (approved):
Changes applied in 8b0091c.

PR #305, opened 2026-08-18
fix(judge): {"match": "false"} is scored as a PASS — parse verdicts as tri-state (closes #295)

### LINKED ISSUE #295: judge: {"match": "false"} is scored as PASS (bool("false") is True), and judge_llm defaults a missing verdict to pass
`src/clawbench/runner/judge.py:187`:

```python
return bool(obj.get("match")), str(obj.get("reason", ""))
```

If a judge model replies `{"match": "false", "reason": "..."}` — a string, which LLMs emit routinely — then `bool("false") is True` and **a mismatch is scored as a PASS**. Verified: `python3 -c 'print(bool("false"))'` → `True`.

`src/clawbench/runner/judge_llm.py:156` has the same bug plus a worse default:

```python
return bool(obj.get("match", True)), str(obj.get("reason", ""))
```

A verdict object missing the `match` key defaults to **True** (pass). This is the lenient rubric — the one that produces the published Reward-lenient column.

Also conflated today: a missing/garbled verdict is indistinguishable from a genuine mismatch, both collapsing to `False` in judge.py.

**Ask:** parse tri-state explicitly — accept real booleans, accept case-insensitive `"true"`/`"false"` strings, return `None` (inconclusive) for anything else, and never default a missing key to a pass. `_parse_verdict` currently has no unit tests; a small matrix (bool / string / missing / garbage / nested) would pin this.

Impact is scoring integrity on published numbers, so worth doing before the next leaderboard refresh.

reacher-z:
Closes #295. Scoring-correctness fix — it changes numbers, so it deserves a careful read.

## The bug

Both judges turned the model's verdict into a boolean with `bool()`:

```python
# judge.py:187 (strict rubric)
return bool(obj.get("match")), str(obj.get("reason", ""))
# judge_llm.py:156 (lenient rubric — produces the published Reward-lenient column)
return bool(obj.get("match", True)), str(obj.get("reason", ""))
```

`bool("false")` is `True`. A judge replying `{"match": "false"}` — a string, which models emit routinely — had its **mismatch recorded as a PASS**. Reproduced on `main` before the fix:

| judge reply | `judge.py` | `judge_llm.py` |
|---|---|---|
| `{"match": "false", "reason": "wrong item"}` | **True** ❌ | **True** ❌ |
| `{"match": "FALSE"}` | **True** ❌ | **True** ❌ |
| `{"reason": "forgot the verdict"}` | False | **True** ❌ |
| `{"match": false}` | False ✓ | False ✓ |

`judge_llm.py` was worse on two counts: a missing `match` key defaulted to `True`, and its keyword fallback returned `True` for *any* reply it could not parse. Both inflate the lenient column specifically.

## The fix

`_coerce_match()` maps real booleans and the common string spellings (`"true"/"false"/"yes"/"no"/"pass"/"fail"/"match"/"mismatch"`, case-insensitive) to a verdict, and everything else — missing key, `null`, `"maybe"`, junk — to `None`.

`None` is not a new state: `judge.py:248/267` already returns it for call failures, and `run.py` gates on `judge_result.get("match") is True` (`:662`, `:746`), so an inconclusive verdict is not a pass. The lenient keyword fallback now returns `None` instead of inventing a pass.

After the fix, every row above resolves correctly (`"false"`/`"FALSE"` → `False`, missing key → `None`).

## Tests

New parse matrix over both modules — booleans, string spellings, casing, fenced JSON, missing key, `null`, non-verdict junk, and a type-rejection case for `_coerce_match`. Full suite: **196 passed**; ruff and pyright clean.

## What this means for published numbers

Any run whose judge answered with a stringly-typed mismatch was scored as a pass. I have not quantified how many rows that affects — it depends on how often each judge model stringifies, which is visible in the stored raw verdicts. **Suggested follow-up:** grep the archived judge outputs for `"match": "` and, if there are hits, re-run `clawbench-rescore` for the affected rows before the next leaderboard refresh. Happy to do that as a separate PR.

@Perry2004 — please review carefully: this changes what counts as a pass. The two calls I'd most like a second opinion on are (1) treating a missing verdict key as inconclusive rather than a mismatch, and (2) the accepted string vocabulary — I kept it deliberately narrow.

### DIFF
--- a/src/clawbench/runner/judge.py
+++ b/src/clawbench/runner/judge.py
@@ -172,6 +172,25 @@ def _call_anthropic_messages(
     return resp["content"][0]["text"]
 
 
+def _coerce_match(value: object) -> bool | None:
+    """Normalize a judge's `match` field into a tri-state verdict.
+
+    Models answer with real booleans, but also with the *strings* "true" and
+    "false", and occasionally with nothing at all. `bool("false")` is True, so
+    a stringly-typed mismatch used to score as a pass; a missing key used to
+    score as a hard mismatch rather than as inconclusive.
+    """
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, str):
+        token = value.strip().strip('".').lower()
+        if token in {"true", "yes", "pass", "match"}:
+            return True
+        if token in {"false", "no", "fail", "mismatch"}:
+            return False
+    return None
+
+
 def _parse_verdict(text: str) -> tuple[bool | None, str]:
     """Best-effort parse of the judge's reply into (match, reason)."""
     text = (text or "").strip()
@@ -184,7 +203,7 @@ def _parse_verdict(text: str) -> tuple[bool | None, str]:
         start = text.index("{")
         end = text.rindex("}") + 1
         obj = json.loads(text[start:end])
-        return bool(obj.get("match")), str(obj.get("reason", ""))
+        return _coerce_match(obj.get("match")), str(obj.get("reason", ""))
     except (ValueError, json.JSONDecodeError):
         # Heuristic fallback
         low = text.lower()
--- a/src/clawbench/runner/judge_llm.py
+++ b/src/clawbench/runner/judge_llm.py
@@ -30,6 +30,8 @@ import urllib.request
 import urllib.error
 from typing import Any
 
+from clawbench.runner.judge import _coerce_match
+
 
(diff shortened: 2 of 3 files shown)


Perry2004 (approved):
I have checked that the bug only happens when the judge LLM returns a malformed result (returning falsy strings/texts instead of valid boolean). Therefore it wont influence the validity of existing judges. 

The fix LGTM. Will merge and include in a patch release.

PR #312, opened 2026-08-20
fix(batch): one truncated run-meta.json no longer discards the batch summary

### LINKED ISSUE #303: One truncated run-meta.json aborts end-of-batch stats before batch-summary.json is written; metadata writes are not atomic
Two cheap robustness gaps around run metadata:

**1. One corrupt `run-meta.json` crashes end-of-batch stats.** `src/clawbench/runner/batch.py:451` does an unguarded `json.loads(meta_file.read_text())` inside `print_run_stats`, which runs at `batch.py:753` — *before* `write_summary_json` at `:754`. So a single malformed metadata file throws after a multi-hour batch, and `batch-summary.json` is never written at all.

**2. The tool can produce that corrupt file itself.** `run_support/metadata.py:286-288` writes with a plain `write_text` (no temp file + rename). A run killed mid-write — and `batch.py:338`/`:701` do SIGKILL on double Ctrl+C — can leave truncated JSON on disk.

Together: interrupt a batch, resume it, and the resumed batch can die at the summary step because of a file the previous interrupt truncated.

**Ask:**
1. `try/except (json.JSONDecodeError, OSError)` per run directory in `print_run_stats` — skip and warn, never abort the summary.
2. Make `write_run_meta` atomic (`NamedTemporaryFile` in the same directory + `os.replace`).

vaibhavdabas16:
## What does this PR do?

Fixes #303: a single truncated `run-meta.json` currently destroys the machine-readable results for an entire batch, and the truncation itself is made impossible in the first place.

**The crash.** `print_run_stats()` parsed every `run-meta.json` with a bare `json.loads()` (`batch.py:450`). `async_main()` calls it at `batch.py:753`, one line *before* `write_summary_json()` at `:754` — so one malformed file raised `JSONDecodeError` out of the function and took `batch-summary.json` down with it. Every other job in the batch lost its recorded results, even though those runs succeeded.

**The truncation.** `write_run_meta()` (`metadata.py:288`) wrote with a plain `write_text()`. A run killed mid-write — Ctrl-C, OOM, host shutdown — leaves a half-written file on disk, which is precisely what the reader above then chokes on.

Both halves are addressed:

- New `clawbench.utils.jsonio` with `write_json_atomic()` (same-directory temp file, flush + `fsync`, then `os.replace`) and `read_json_or_none()`. `os.replace` replaces atomically on POSIX and Windows, so an interrupted write leaves either the previous file or no file — never a truncated one.
- `write_run_meta()` and `write_summary_json()` now write atomically.
- `print_run_stats()` reads via `read_json_or_none()`. It also rejects valid JSON that isn't an object — a bare list previously reached `.get()` and raised `AttributeError` — prints a warning naming the offending file, and falls back to the directory name for that one run.

Reporting now degrades to a single warning line for the affected run instead of losing the whole batch.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

<sub>Host-side runner/reporting change; no task data involved.</sub>

## Test plan

- [x] New `tests/test_batch_stats_resilience.py` (11 tests): atomic round-trip and parent-dir creation, no leftover temp files, previous file survives a failed serialization, `read_json_or_none()` against truncated / empty / garbage / undecodable input, and two `print_run_stats()` regression tests covering a truncated `run-meta.json` and a non-object one.
- [x] Confirmed the regression tests genuinely catch the bug: reverted `batch.py` and `metadata.py` to `main` and re-ran, which fails with `JSONDecodeError` and `AttributeError` respectively; both pass with the fix applied.
- [x] Full suite: `203 passed, 3 skipped`. The one unrelated failure, `test_host_tasks.py::test_checked_task_json_files_parse_and_validate[v1-lite]`, reproduces identically on a clean `main` checkout on this machine — the `v1-lite` task files are git symlinks (mode `120000`) that Windows checks out as plain text.
- [x] `ruff check` and `ruff format --check` clean on all changed files. Removing the last `json.` uses from `batch.py` left `import json` unused, so it is dropped (F401).

## Related issues

Fixes #303.

<sub>One note for maintainers, deliberately left out of scope: `eval/rescore.py` reads `run-meta.json` with the same unguarded `json.loads()` at lines 58, 102 and 279. Happy to follow up in a separate PR if that is wanted.</sub>

### DIFF
--- a/src/clawbench/runner/batch.py
+++ b/src/clawbench/runner/batch.py
@@ -4,7 +4,6 @@ import argparse
 import asyncio
 import fnmatch
 import itertools
-import json
 import os
 import re
 import shutil
@@ -17,6 +16,7 @@ from pathlib import Path
 
 import yaml
 
+from clawbench.utils.jsonio import read_json_or_none, write_json_atomic
 from clawbench.utils.paths import ASSET_ROOT, WORKSPACE_ROOT, ensure_workspace_templates
 
 
@@ -447,8 +447,15 @@ def print_run_stats(base_output: Path) -> None:
 
             # Parse case and model from run-meta.json or dir name
             meta_file = run_dir / "run-meta.json"
-            if meta_file.exists():
-                meta = json.loads(meta_file.read_text())
+            meta = read_json_or_none(meta_file) if meta_file.exists() else None
+            if meta is not None and not isinstance(meta, dict):
+                meta = None
+            if meta is None and meta_file.exists():
+                print(
+                    f"  WARNING: unreadable {meta_file}; "
+                    "falling back to directory name for this run"
+                )
+            if meta is not None:
                 case = meta.get("test_case", "?")
                 model = meta.get("model", model_dir.name)
                 intercepted = meta.get("intercepted", False)
@@ -576,7 +583,7 @@ def write_summary_json(
             for s in ("passed", "failed", "error", "skipped")
         },
     }
-    (base_output / "batch-summary.json").write_text(json.dumps(data, indent=2))
+    write_json_atomic(base_output / "batch-summary.json", data)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Perry2004:
Thanks for the contribution. I'll take a look.

Perry2004:
I believe it doesn't worth adding such amount of custom atomic disk io snippet just to handle such rare case. 
Rather I'd suggest to change the order to put `print_run_stats` as the last step so even it fails there's no data loss. 

Closed in favor of #325.

PR #313, opened 2026-08-20
fix(rescore): resolve paths from the workspace instead of a maintainer's home

### LINKED ISSUE #296: rescore: public reproducibility CLI defaults to a maintainer's home paths and silently no-ops elsewhere
`src/clawbench/eval/rescore.py:203-217` ships a maintainer's personal machine layout as the public defaults:

```python
p.add_argument("--sweep-root",   type=Path, default=Path.home() / "work/ClawBench/claw-output/sweep")
p.add_argument("--models-yaml",  type=Path, default=Path.home() / "work/ClawBench/models/models.yaml")
```

`clawbench-rescore` is an installed console script (`pyproject.toml:29`) and both `docs/scoring.md` and `eval/scoring.md` advertise it as *the* way to reproduce every leaderboard number from public traces. On any machine that isn't that one, `--sweep-root` doesn't exist, the rglob matches nothing, and the tool prints "discovered 0 tasks" and **exits 0** — a silent no-op for the exact audience we most want to succeed.

Every other code path resolves paths through `WORKSPACE_ROOT` / `MODELS_YAML` (`run_support/config.py:91`); rescore bypasses that.

**Ask:**
1. Default `--models-yaml` to the workspace-resolved `config.MODELS_YAML`.
2. Default `--sweep-root` to `./test-output` (where runs actually land) or require it explicitly.
3. Error loudly when the root does not exist, instead of reporting 0 tasks and exiting 0.
4. While there: rescore only accepts the `api_key` scalar form, rejecting the `api_keys` list form that `load_model_config` (`config.py:192-196`) normalizes — reuse `load_model_config` instead of re-parsing.

vaibhavdabas16:
## What does this PR do?

Fixes #296 — all four asks in the issue.

`clawbench-rescore` is an installed console script (`pyproject.toml:29`) that `docs/scoring.md` and `eval/scoring.md` both advertise as *the* way to reproduce leaderboard numbers from public traces. It shipped one machine's layout as its defaults:

```python
--sweep-root   ~/work/ClawBench/claw-output/sweep
--models-yaml  ~/work/ClawBench/models/models.yaml
```

`find_run_dirs()` reaches the root through `Path.rglob()`, which yields nothing for a directory that does not exist rather than raising. So on any other machine the tool printed `discovered 0 tasks` and **exited 0** — a silent no-op for precisely the audience the script exists to serve.

| Ask | Change |
| --- | --- |
| 1. Default `--models-yaml` to the workspace-resolved config | Now defaults to `MODELS_YAML` |
| 2. Default `--sweep-root` to where runs land | Now `WORKSPACE_ROOT / "test-output"`, matching `clawbench-batch --output-dir` (`batch.py:833`) |
| 3. Error loudly instead of no-opping | A `--sweep-root`/`--only-batch` that isn't a directory, or a tree with no `run-meta.json` under it, prints a diagnostic naming the path and exits **2** |
| 4. Reuse `load_model_config` | The judge model is resolved through the shared loader instead of a private `yaml.safe_load()` + `config.get("api_key")` |

Before and after, on a path that doesn't exist:

```console
$ clawbench-rescore --sweep-root /definitely/not/here     # before
discovered 0 tasks, judging 0 intercepted ones ...
$ echo $?
0

$ clawbench-rescore --sweep-root /definitely/not/here     # after
ERROR: --sweep-root /definitely/not/here is not a directory.
       Point it at a tree containing completed runs (the runner writes them under ./test-output by default).
$ echo $?
2
```

### One structural change worth calling out

Ask 4 could not be satisfied by importing `run_support/config.py`: that module runs `ENGINE = _detect_engine()` at import time (`config.py:90`) and `sys.exit(1)`s when neither Docker nor Podman is on `PATH`. `rescore.py` and both judges currently import none of it, so `clawbench-rescore` runs fine on a host with no container runtime — which is the normal case for someone scoring published traces. Importing `config` to reuse the loader would have handed a post-hoc scoring tool a hard Docker dependency, making the portability problem worse.

So `load_models_yaml()` and `load_model_config()` move to `utils/model_config.py` — they only ever needed `WORKSPACE_ROOT` and `yaml` — and both gain an optional explicit path argument. `run_support/config.py` re-exports both names (they were already in its `__all__`), so every existing importer, `runner/run.py` included, is untouched. A test asserts `rescore` never reaches for `run_support.config` again.

Happy to split that move into its own commit or drop it in favour of a narrower fix if you'd rather keep the loaders where they are.

Also extracted the argument parser into `build_parser()` so the shipped defaults are assertable from tests.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

<sub>Host-side eval tooling; no task data involved.</sub>

## Test plan

- [x] New `tests/test_rescore_cli.py` (10 tests): both defaults resolve under `WORKSPACE_ROOT`; a regression guard asserting no shipped default contains `work/ClawBench`; exit code 2 with a useful message for a missing root, an existing-but-empty root, and a missing `--only-batch`; `load_model_config` accepting both the `api_keys` list form and the `api_key` scalar form; the error naming the explicit `--models-yaml` path; and a guard that `rescore` does not import the container-probing module.
- [x] Verified the old defaults really were `Path.home() / "work/ClawBench/..."` by reverting the file and re-reading them, and confirmed `Path("/definitely/not/here").rglob("run-meta.json")` returns `[]` rather than raising — the mechanism behind the silent no-op.
- [x] Confirmed `import clawbench.eval.rescore` succeeds with `shutil.which` stubbed to return `None` for every command, i.e. with no container engine installed at all.
- [x] Full suite: `202 passed, 3 skipped`. The single failure, `test_host_tasks.py::test_checked_task_json_files_parse_and_validate[v1-lite]`, reproduces identically on a clean `main` on this machine — the `v1-lite` task files are git symlinks (mode `120000`) that Windows checks out as text. Unrelated to this change.
- [x] `ruff check` and `ruff format --check` clean on all four files.

## Related issues

Fixes #296.

<sub>Noticed while in here, left out of scope: `batch.py:44-61` and `tui.py:110`/`:346` each carry their own private copy of `MODELS_YAML` + `load_models_yaml`. Now that there is one shared home for them, those two could be collapsed onto it in a follow-up if you want it.</sub>

### DIFF
--- a/src/clawbench/eval/rescore.py
+++ b/src/clawbench/eval/rescore.py
@@ -36,7 +36,12 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 from typing import Any
 
-import yaml
+from clawbench.utils.model_config import (
+    MODELS_YAML,
+    ModelConfigError,
+    load_model_config,
+)
+from clawbench.utils.paths import WORKSPACE_ROOT
 
 JUDGE_FILE = {"strict": "judge.json", "lenient": "judge_llm.json"}
 
@@ -199,14 +204,16 @@ def write_eval_results(
     print(f"  eval_results written to {out_dir}/")
 
 
-def main() -> int:
+def build_parser() -> argparse.ArgumentParser:
     p = argparse.ArgumentParser(
         description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
     )
     p.add_argument(
         "--sweep-root",
         type=Path,
-        default=Path.home() / "work/ClawBench/claw-output/sweep",
+        default=WORKSPACE_ROOT / "test-output",
+        help="Directory tree to scan for completed runs "
+        "(default ./test-output, where the runner writes them).",
     )
     p.add_argument(
         "--judge-model",
@@ -217,7 +224,8 @@ def main() -> int:
     p.add_argument(
         "--models-yaml",
         type=Path,
-        default=Path.home() / "work/ClawBench/models/models.yaml",
+        default=MODELS_YAML,
+        help=f"Model definitions to resolve --judge-model against (default {MODELS_YAML}).",
     )
     p.add_argument(
         "--rubric",
@@ -245,19 +253,23 @@ def main() -> int:
     )
     p.add_argument("--limit", type=int, default=0)
     p.add_argument("--only-batch", type=Path, default=None)
-    args = p.parse_args()
+    return p
 
-    cfg_all = yaml.safe_load(args.models_yaml.read_text())
-    if args.judge_model not in cfg_all:
-        print(
-            f"ERROR: judge model {args.judge_model!r} not in {args.models_yaml}",
-            file=sys.stderr,
-        )
-        return 2
-    judge_cfg = dict(cfg_all[args.judge_model])
-    if not judge_cfg.get("api_key"):
-        print(f"ERROR: judge {args.judge_model!r} has no api_key", file=sys.stderr)
-        return 2
+
+def main() -> int:
+    args = build_parser().parse_args()
+
+    # Shared loader: validates the model exists, checks required fields, and
+    # normalizes the api_key/api_keys pair. Re-parsing the YAML here used to
+    # reject the api_keys list form that every other entry point accepts.
+    try:
+        judge_cfg = load_model_config(args.judge_model, args.models_yaml)
+    except ModelConfigError as e:
+        # A bad --judge-model is a typo on the command line, caught before any
+        # scoring work starts. Report it like run.py does rather than letting
+        # the traceback out.
+        print(f"ERROR: {e}")
+        return 1
 
     rubrics = [args.rubric] if args.rubric != "both" else ["lenient", "strict"]
     judge_funcs = {}
@@ -270,11 +282,25 @@ def main() -> int:
 
         judge_funcs["lenient"] = judge_lenient
 
-    run_dirs = (
-        find_run_dirs(args.only_batch)
-        if args.only_batch
-        else find_run_dirs(args.sweep_root)
-    )
+    scan_root = args.only_batch or args.sweep_root
+    if not scan_root.is_dir():
+        flag = "--only-batch" if args.only_batch else "--sweep-root"
+        print(f"ERROR: {flag} {scan_root} is not a directory.", file=sys.stderr)
+        print(
+            "       Point it at a tree containing completed runs "
+            "(the runner writes them under ./test-output by default).",
+            file=sys.stderr,
+        )
+        return 2
+
+    run_dirs = find_run_dirs(scan_root)
+    if not run_dirs:
+        print(
+            f"ERROR: no runs found under {scan_root} "
+            "(looked for run-meta.json at any depth).",
+            file=sys.stderr,
+        )
+        return 2
 
     pending = []
     for rd in run_dirs:
--- a/src/clawbench/runner/run_support/config.py
+++ b/src/clawbench/runner/run_support/config.py
@@ -7,13 +7,17 @@ import sys
 import warnings
 from pathlib import Path
 
-import yaml
-
 from clawbench.runner.run_support.harness_registry import (
     HARNESS_REGISTRY,
     HARNESS_REGISTRY_YAML,
     HarnessRegistry,
     load_harness_registry,
+)
+from clawbench.utils.model_config import (
+    MODELS_YAML,
+    ModelConfigError,
+    load_model_config,
+    load_models_yaml,
 )
 from clawbench.utils.paths import (
     ASSET_ROOT,
@@ -45,17 +49,6 @@ __all__ = [
     "resolve_test_case_dir",
     "resolve_test_case_path",
 ]
-
-
-class ModelConfigError(Exception):
-    """Raised when a model config in models/models.yaml is missing or invalid.
-
-    A plain Exception (not SystemExit) so callers that load a model mid-run
-    (e.g. the judge stage, after the agent has already produced results) can
-    catch it and continue instead of the process dying before run-meta.json
-    is written.
-    """
-
 
 HARNESSES = HARNESS_REGISTRY.harnesses
 DEFAULT_HARNESS = HARNESS_REGISTRY.default
@@ -126,9 +119,6 @@ def __getattr__(name: str) -> str:
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
-MODELS_YAML = WORKSPACE_ROOT / "models" / "models.yaml"
-
-
 def load_dotenv(path: Path) -> dict[str, str]:
     env = {}
     if not path.exists():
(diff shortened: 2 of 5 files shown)


Perry2004 (changes requested):
At least one of the added test case is failing in CI. Please resolve this before further review, thanks. 
https://github.com/TIGER-AI-Lab/ClawBench/actions/runs/32335738442/job/96648767983?pr=313

vaibhavdabas16:
Pushed `431c409` to fix the red CI on this PR. The failure was in a test I added here, not in the fix itself.

### What failed

`tests/test_rescore_cli.py::test_defaults_do_not_ship_a_maintainers_home_layout` failed on `ubuntu-latest` and `macos-latest` and passed on `windows-latest`:

```
AssertionError: assert 'work/ClawBench' not in '/home/runner/work/ClawBench/ClawBench/test-output'
```

The guard was written as a substring check against the *resolved* default paths:

```python
for value in (args.sweep_root, args.models_yaml):
    assert "work/ClawBench" not in value.as_posix()
```

Actions checks this repo out to `/home/runner/work/ClawBench/ClawBench`, so a default that is correctly derived from `WORKSPACE_ROOT` still contains the substring `work/ClawBench`. The Windows runner passed only because its checkout path is `D:\a\ClawBench\ClawBench`. The assertion was testing the runner's directory layout, not the code.

### What changed

The property actually worth guarding is that the defaults are not anchored to the invoking user's home directory. That cannot be read off the resolved paths from an arbitrary checkout location, so it is now asserted against the source of `build_parser()`:

```python
src = inspect.getsource(rescore.build_parser)

assert "Path.home()" not in src
assert "expanduser" not in src
```

`test_defaults_are_workspace_relative` already pins the resolved values exactly (`WORKSPACE_ROOT / "test-output"` and `MODELS_YAML`), so coverage of #296 is unchanged. The diff is test-only — no change to `rescore.py`, `model_config.py`, or `config.py`.

### Verification

I reproduced the CI condition locally by forcing the workspace to a runner-shaped path via `CLAWBENCH_WORKSPACE=.../home/runner/work/ClawBench/ClawBench`:

| | under a `work/ClawBench` workspace |
|---|---|
| old assertion | fails, with the same pytest diff as CI |
| new assertion | 10 passed |

The control matters: the old form fails under that path, so the reproduction is faithful rather than a vacuous pass.

Full suite locally: 202 passed, 3 skipped. The one failure, `test_host_tasks.py::test_checked_task_json_files_parse_and_validate[v1-lite]`, is a local Windows artifact — those task files are git symlinks (mode `120000`) that Windows checks out as text — and it is green on CI.

### Current state

The workflow runs for the new head are sitting in `action_required` (0s, never executed), so the checks on this PR are pending a maintainer approving the run rather than failing. Happy to adjust the guard if you would prefer a different form.

vaibhavdabas16:
### Note for maintainers: this PR overlaps #314

Separate from the CI fix above, and only relevant at merge time.

This PR and #314 both touch `src/clawbench/runner/run_support/config.py`, from opposite directions:

- **#313 (this PR)** moves `load_models_yaml()` and `load_model_config()` out to `utils/model_config.py` and leaves re-exports behind, so that `eval/rescore.py` can reuse them. It cannot import `run_support/config.py` directly: that module resolves a container engine at import time and exits when neither Docker nor Podman is present, which would make a post-hoc scoring tool require a container runtime.
- **#314** rewrites those same two functions in place, so they raise `ModelConfigError` instead of calling `sys.exit`.

Each merges cleanly into `main` on its own. Verified against the current heads (`431c409` and `2c777ab`) with `git merge-tree`:

```
313 into main  -> clean
314 into main  -> clean
313 vs 314     -> CONFLICT (content): src/clawbench/runner/run_support/config.py
```

**So two things can happen, depending on merge order:**

1. **#313 lands first** — #314 then needs a rebase, and its `ModelConfigError` change should be applied to the moved functions in `utils/model_config.py` rather than to `run_support/config.py`. The re-exports mean nothing else has to move.
2. **#314 lands first** — #313 then needs a rebase, carrying #314's `ModelConfigError` bodies through the move into `utils/model_config.py`.

Either order works and neither is blocked; only the second one to merge needs the rebase. No preference from me — say which you would rather take first and I will rebase the other onto it promptly.

One thing worth a maintainer's eye either way: after both land, `load_model_config()` will raise `ModelConfigError` in a path that `rescore` calls. This PR's `test_load_model_config_reports_the_explicit_path_on_a_bad_model` currently asserts `SystemExit`, and that assertion will need to change to match whichever behaviour ends up canonical.

vaibhavdabas16:
Rebased onto `main` at `890e0be`. #314 landed first, so this took option 2 from my earlier merge-order note: #314's `ModelConfigError` bodies are carried through the move rather than reverted.

### What the conflict was

#314 rewrote `load_models_yaml()` and `load_model_config()` in place in `run_support/config.py` to raise `ModelConfigError` instead of calling `sys.exit`. Those are the same two functions this PR moves out to `utils/model_config.py`, which is why the two conflicted.

### How it resolved

- `ModelConfigError` is now defined in `utils/model_config.py`, beside the loaders it guards, and re-exported from `run_support/config.py` along with `MODELS_YAML` and both loaders. `run.py`'s two `except ModelConfigError` handlers and `config.__all__` are unchanged.
- The moved loaders keep both #314's raising behaviour and this PR's optional `models_yaml: Path | None` override.

### Two things that fell out of the merge

**1. `rescore.main()` would have leaked a traceback.** It called `load_model_config` bare, relying on the old `sys.exit`. With #314's behaviour a bad `--judge-model` — a command-line typo, caught before any scoring work starts — would print a traceback instead of the one-line `ERROR:` it used to. It now catches and returns 1, matching how `run.py` handles the same error. New test: `test_rescore_reports_a_bad_judge_model_instead_of_a_traceback`.

**2. One of #314's tests needed its patch target moved.** `test_run_judge_stage.py::test_missing_models_yaml_is_also_catchable` patched `MODELS_YAML` on `run_support.config`. After the move that module only re-exports the name, so the patch no longer reached the loaders and the test failed:

```
assert 'absent.yaml' in str(excinfo.value)
AssertionError: assert 'absent.yaml' in '.../models/models.yaml not found (copy models.example.yaml and fill in your keys)'
```

It patches `clawbench.utils.model_config` instead. The property under test — a missing `models.yaml` raises `ModelConfigError` rather than `SystemExit`, through both `load_models_yaml()` and `load_model_config()` — is unchanged; only the module the patch lands on moved. This is the assertion change I flagged as needed in my earlier note.

As predicted there, `test_load_model_config_reports_the_explicit_path_on_a_bad_model` also changed: it asserted `SystemExit` plus captured stdout, and now asserts `ModelConfigError` and that the message names the explicit path rather than the workspace default.

### Verification

- CI green on `890e0be`: `pytest` on ubuntu / macos / windows, `static-check`, `merge-gatekeeper`. The ubuntu and macos runs are where the original failure on this PR was.
- Locally 221 passed, 4 skipped, 4 deselected. The 4 deselected are `test_host_tasks.py::test_checked_task_json_files_parse_and_validate`, which I excluded because it fails on this Windows checkout only — the task files are git symlinks (mode `120000`) that Windows checks out as text; it is green on CI. The 4 skips are the pre-existing POSIX-only Harbor ones in `test_harbor_correctness.py` and `test_prorl_mock.py`, unrelated to this PR.
- `ruff check` on the touched files reports exactly the same findings as those files on `main` — nothing new introduced.
- `git merge-tree` against `main`: clean.

@Perry2004 this should be ready for another look whenever you have time — the red CI you flagged was fixed in `431c409` and the conflicts are gone.

Perry2004:
Thanks, I'll take a look.

vaibhavdabas16:
Rebased onto `main` at `9dd9d44` (#339 made the engine probe lazy, so the only conflict was around `config.ENGINE`; `utils/model_config.py`'s docstring no longer cites the import-time probe as the reason for the split). CI green.

Issue #315, opened 2026-08-20
run_support/config.py resolves a container engine at import time, so any module importing it inherits a hard Docker dependency


vaibhavdabas16:
# run_support/config.py resolves a container engine at import time, so any module that imports it inherits a hard Docker dependency

`src/clawbench/runner/run_support/config.py:90` runs engine detection as a side effect of import:

```python
ENGINE = _detect_engine()
```

and `_detect_engine` (`config.py:83-87`) exits the process when neither engine is present:

```python
for cmd in ("docker", "podman"):
    if shutil.which(cmd):
        return cmd
print("ERROR: Neither 'podman' nor 'docker' found on PATH")
sys.exit(1)
```

Because this happens at module scope, `import clawbench.runner.run_support.config` — or anything that transitively imports it, including `run_support.docker` and `run_support.metadata` — terminates the interpreter on a host with no container runtime. Not an exception that a caller can catch and degrade on: a bare `sys.exit`.

## Why this matters beyond tidiness

**It has already forced workarounds in seven test files.** Every one of these monkeypatches `shutil.which` for no reason other than getting past the probe:

```
tests/test_cli_entrypoints.py       tests/test_harbor_adapter.py
tests/test_mock_container_runtime.py tests/test_model_api_preflight.py
tests/test_results_and_metadata.py   tests/test_tui_helpers.py
tests/test_host_timeout.py
```

A typical instance, from `test_results_and_metadata.py:179`:

```python
monkeypatch.setattr(shutil, "which", lambda cmd: cmd)
metadata = importlib.import_module("clawbench.runner.run_support.metadata")
```

Tests that never touch a container still have to fake one into existence to import the module under test.

**It shapes production code too.** Two examples from the current tree:

- `eval/rescore.py` scores runs that have already finished and needs nothing from a container engine. It deliberately does not import `run_support.config`, which is why it grew its own `yaml.safe_load` of `models.yaml` instead of reusing `load_model_config` — the divergence reported in #296.
- `runner/batch.py` imports cleanly today with no engine present, and `tests/test_batch_and_tui_helpers.py` depends on that. Any constant or helper it wants from `run_support.docker` has to be duplicated or relocated instead of imported.

Both are avoidable. The engine is a runtime dependency of *executing* a container, not of *importing* the module that knows how.

## Suggested fix

Make resolution lazy and cached:

```python
@functools.lru_cache(maxsize=1)
def engine() -> str:
    ...  # current _detect_engine body
[...]

Issue #302, opened 2026-08-18
A missing judge model kills the process after the agent run and loses run-meta.json entirely (SystemExit escapes both handlers)


reacher-z:
`src/clawbench/runner/run.py:578` loads the judge config **after** the agent run finishes, inside the judge stage:

```python
judge_cfg = load_model_config(args.judge)
```

But `load_model_config` calls `sys.exit(1)` on a missing model (`run_support/config.py:164-166`, also `:180`, `:191`, `:200`). `SystemExit` derives from `BaseException`, so it escapes both the judge-stage `except Exception` (`run.py:606`) and the outer `except Exception` (`run.py:674`) — and `write_run_meta` (`run.py:665`/`:722`) never runs.

Net effect for a first-time user whose `models.yaml` lacks `deepseek-v4-pro` (the default judge — `models.example.yaml` even warns it "must exist here or the scoring stage fails"): they burn a full agent run of up to 30 minutes, the interception may well have succeeded, and then the process dies leaving **no `run-meta.json` at all**. The run is invisible to batch stats, to `clawbench-rescore` (which globs `run-meta.json`), and to HF upload.

**Ask:**
1. Validate the judge config at startup, before the agent run — next to the existing `preflight_model_api` call at `run.py:352`.
2. Make `load_model_config` raise a catchable exception rather than `sys.exit`, or catch `SystemExit` in the judge stage and degrade to a `judge_setup_failed` outcome that still writes `run-meta.json`.

Either way the invariant worth stating: **a completed agent run always writes its metadata**, whatever happens at scoring time.

reacher-z:
@Perry2004 — filed from a read-only audit of `main`; every claim is cited to file:line and was re-verified by hand before filing. Flagging for your triage: happy to send a PR for any of these you want picked up, or to close it if you judge the behaviour intentional.

PR #316, opened 2026-08-20
fix(runner): bound container wait so one wedged run can't stall a batch

### LINKED ISSUE #298: No host-side timeout on container wait — one wedged container stalls an entire batch indefinitely
There is no host-side deadline on a run.

- `src/clawbench/runner/run_support/docker.py:522-585` — `docker_wait` blocks on `[ENGINE, "wait", name]` with no timeout.
- `src/clawbench/runner/batch.py:313` — `await proc.communicate()` has no per-job timeout.

The only time limit lives *inside* the container (`runtime/harnesses/base/entrypoint.sh:251`, `MAX_WAIT=${TIME_LIMIT_S:-1800}`). If that watchdog never fires — entrypoint crash, wedged Chromium, container engine hiccup, zombie container — the host waits forever: `clawbench-run` blocks, and in batch mode the job holds a concurrency slot indefinitely. A 130-task batch can stall overnight with no error and no summary.

Given batches routinely run 8–20 hours (per `wall_hours` in the leaderboard CSV), a single wedged container is an expensive failure mode.

**Ask:**
1. Wrap `docker_wait` in a host-side deadline of `time_limit_s + grace` (e.g. +5 min), then `docker kill` the container and classify the run as `infra_failure` / `host_timeout`.
2. Wrap `proc.communicate()` in `asyncio.wait_for` with the same bound in `batch.py`, marking the job `error: host_timeout` so the batch proceeds.

vaibhavdabas16:
## What does this PR do?

Fixes #298 — both asks.

The only time limit lived *inside* the container (`entrypoint.sh:251`, `MAX_WAIT=${TIME_LIMIT_S:-1800}`). On the host side there was no deadline anywhere:

- `docker_wait()` (`docker.py:522`) blocked on `<engine> wait` with `while proc.poll() is None:` and no bound
- `batch.py:313` awaited `proc.communicate()` with no per-job bound

If the in-container watchdog never fired — entrypoint crash, wedged Chromium, engine hiccup, zombie container — the host waited forever. `clawbench-run` blocked; in batch mode the job held a concurrency slot indefinitely. With batches routinely running 8–20 hours, one wedged container could stall a 130-task run overnight with no error and no summary.

### Ask 1 — deadline in `docker_wait`

It now takes `timeout_s` and returns whether the deadline expired. On expiry it kills the container and returns `True`. `run.py` passes `time_limit_s + HOST_TIMEOUT_GRACE_S` (5 min), records `host_timeout: ...` as the failure reason, and classifies the run as `infra_failure` so it stays out of adjusted scoring — then **carries on** to copy results and write `run-meta.json` as usual. Partial data from a killed container is still worth keeping, and this preserves the same invariant #303 and #302 are about: a run that got far enough to produce anything always records it. `--human` runs stay unbounded by design.

### Ask 2 — per-job bound in `batch.py`

`proc.communicate()` is wrapped in `asyncio.wait_for`, sized from the case's own `time_limit` plus `BATCH_JOB_GRACE_S` (15 min) so it sits *above* the run's own deadline — `clawbench-run` reports its own timeout first, and this only fires when the child process itself is wedged. On expiry it kills the process group, marks the job `error`, notes `host_timeout` in the job log, and the batch proceeds. `--job-timeout` overrides the derived value; `0` disables it.

`job_timeout_s()` reads `time_limit` from both corpus layouts — `<case>/task.json` and claw-eval's flat `<suite>/<id>.json` — and falls back to 1800s when the task file is unreadable.

### On the new `utils/timeouts.py`

Three constants shared by `run.py`, `docker.py` and `batch.py`. `batch.py` cannot import them from `run_support.docker`, because that pulls in `run_support.config`, which resolves a container engine at import time and `sys.exit`s when neither Docker nor Podman is installed — and `batch.py` must stay importable without one (`tests/test_batch_and_tui_helpers.py` depends on it).

A shared module seemed better than duplicating two integers across modules that must agree. **That import-time probe is the underlying problem, and I've filed it separately as #315** with a suggested lazy-`engine()` fix; if that lands, this module could fold back into `docker.py`. Happy to take a different shape here if you'd prefer.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

<sub>Host-side runner/batch change; no task data involved.</sub>

## Test plan

- [x] New `tests/test_host_timeout.py`, 12 tests: `job_timeout_s` against both corpus layouts, four unreadable-task fallbacks, override and disable; `docker_wait` killing a never-exiting container and returning promptly; `timeout_s=None` preserving the old unbounded behaviour; `BATCH_JOB_GRACE_S > HOST_TIMEOUT_GRACE_S` so the layering cannot silently invert; and `batch.py` still importing with no container engine present.
- [x] Demonstrated the hang directly rather than only through tests — driving the **unpatched** `docker_wait` against a stub container that never exits, on a background thread:
      ```
      UNPATCHED docker_wait still running after 6s? True   <- hangs forever
      PATCHED   returned True after 2.0s; issued: ['kill', 'wedged']
      ```
- [x] Full suite: `204 passed, 3 skipped`. The one failure, `test_host_tasks.py::test_checked_task_json_files_parse_and_validate[v1-lite]`, reproduces identically on a clean `main` on this machine — those task files are git symlinks (mode `120000`) that Windows checks out as text. Unrelated.
- [x] `ruff check` and `ruff format --check` clean.

**Not verified:** no live containerized run — I don't have Docker on this machine. The wedged-container path is exercised by injection, not by a genuinely stuck Chromium.

## Related issues

Fixes #298. Root cause of the `utils/timeouts.py` workaround filed as #315.

<sub>One deliberate choice worth a maintainer's eye: a host timeout is classified as the existing `infra_failure` category with `host_timeout: ...` in `failure_reason`, matching the convention at `run.py:415`. The issue's wording ("infra_failure / host_timeout") could instead mean a distinct `host_timeout` category — that would need adding it to `NON_MODEL_FAILURE_CATEGORIES` and widening the `infra_failure` predicate in `classify_run`, which changes semantics shared with every other failure path. I took the conservative route; say the word and I'll switch it.</sub>


src/clawbench/runner/batch.py
```diff
@@ -309,17 +328,47 @@ async def run_job(
...
+                except asyncio.TimeoutError:
+                    # The child is wedged past even its own host-side deadline.
+                    # Kill the whole process group and keep the batch moving.
+                    host_timed_out = True
+                    print(
+                        f"[{ts()}] HOST TIMEOUT {job.case_name} ({job.model}) "
+                        f"after {int(bound or 0)}s — killing"
+                    )
+                    try:
+                        os.killpg(proc.pid, signal.SIGKILL)
```
Perry2004: I think this kills the per-run Python process and left the container process. 
Also it is using SIGKILL so it would bypass the per-run cleanups in the `finally` block.

Perry2004 (changes requested):

src/clawbench/runner/batch.py
vaibhavdabas16: You're right on both counts, and the second one made the first worse. Fixed in `c8ebd8a`.

The container isn't in that process group. `run.py` shells out to the engine (`docker_run(container, ...)`, run.py:508) and the daemon owns the container from there, so `killpg` reached the Python child and the CLI client and nothing else. And because SIGKILL can't be caught, `run.py` never reached its `finally: docker_rm(container)` (run.py:743-745, which is an `rm -f`). So every timed-out job leaked exactly the container whose CPU and memory the bound exists to reclaim. The marker I wrote into the job log — `"container and child process killed"` — was describing something the code didn't do.

### The fix

`stop_wedged_job()` sends SIGTERM first. `run.py` already installs a SIGTERM handler that raises `KeyboardInterrupt` (run.py:423-429) and unwinds through that same `finally`, so the container actually goes away. Worth noting that handler is installed for **every** run, not just `--human` — I'd initially misremembered the `--human` Ctrl+C path at run.py:478 as the only graceful hook.

SIGKILL follows only if the child is still alive after `JOB_KILL_GRACE_S` (60s). Graceful teardown shouldn't become a second way to hang the batch, and a run wedged badly enough to ignore SIGTERM must not keep its concurrency slot. The grace covers `docker rm -f`, the disposable-mailbox delete and the browser-runtime teardown — short subprocess and network calls, not agent work.

### Two things I changed beyond the literal comment

- The job log now reports which of the two actually happened. On escalation it says the container may still be running rather than claiming a clean teardown, since that's precisely the case where `run.py`'s cleanup was skipped.
- An already-gone process group is reported as *not* escalated: it means `clawbench-run` reached the end of its own teardown, and labelling it SIGKILLed would send someone hunting for a container that isn't there.

### Tests

Four new tests in `tests/test_host_timeout.py` drive a stub run that honours only the signals it's told to, so the SIGTERM-then-SIGKILL sequence is observable. As a control I reverted `stop_wedged_job` to the old SIGKILL-first behaviour and re-ran — **3 of the 4 fail**, so they're pinning the fix rather than passing vacuously.

One limitation to flag: those 4 skip on Windows, where `signal.SIGKILL` and `os.killpg` don't exist — which is also why this path doesn't run there. I verified them locally by re-running the file with `signal.SIGKILL` shimmed in (16 passed), and that's how the control above was run too, but genuine POSIX coverage comes from the ubuntu and macos jobs here.

Also rebased onto current `main` (the branch was 32 commits behind). Full suite locally: 222 passed, 8 skipped. `ruff format` clean; `ruff check` adds no new finding classes over the previous head.

**Still not verified:** no live containerized run — I don't have Docker on this machine, so the orphaned-container behaviour is reasoned from the code and exercised by injection, not observed against a genuinely stuck Chromium. If you have a box where you can wedge one, that's the check I can't do.

Perry2004:
Resolved conflict with main.

Perry2004 (approved):
LGTM🫡
Thanks for the contribution.

vaibhavdabas16:
thanks @Perry2004 :)

PR #320, opened 2026-08-21
fix(eval): one Stage-1 matching predicate for the live and offline paths

### LINKED ISSUE #301: Stage-1 interceptor matching is duplicated between runtime-server and edgebench_judge, and the live copy has zero tests
The Stage-1 interceptor decision — the benchmark's deterministic ground truth — exists in two hand-maintained copies:

- `src/clawbench/runtime/runtime-server/server.py:173-185` (`_const_fields_match`) and `:449-477` (url / method / body / params gate) — the real thing, running in-container.
- `src/clawbench/eval/edgebench_judge.py:55-94` — a re-implementation whose comment says it "mirrors runtime-server".

Only the mirror has tests (`tests/test_edgebench_judge.py`). `server.py` is 878 lines with **zero** tests and is excluded from pyright (`pyproject.toml:74` excludes `src/clawbench/runtime`). So the copy that decides every published Intercepted number is the unverified one, and any drift between the two means offline judging disagrees with what actually happened during runs.

**Ask:**
1. Extract the matching predicate into a shared module (e.g. `runtime/shared/matching.py`) imported by both.
2. Add a parity test over a fixture matrix of `url_pattern` / `method` / `body` / `params` cases asserting both paths agree — including the regex edge cases from #258 (empty pattern matches everything; malformed pattern raises inside the CDP loop and silently kills interception for the rest of the run).

vaibhavdabas16:
## What does this PR do?

Stage-1 — does this request hit the task's target (`url_pattern` + `method` + const `body`/`params`)? — is the benchmark's deterministic ground truth. It sets `run-meta.intercepted` and every published Intercepted number. It existed as two hand-maintained copies: `runtime-server/server.py:173-185,449-477` running in-container, and a mirror at `eval/edgebench_judge.py:55-94` that re-verifies a submitted evidence archive. Only the mirror had tests, and `server.py` is excluded from pyright (`pyproject.toml:74`) — so the copy deciding the published numbers was the unverified one.

They had drifted, in two ways that change verdicts.

**1. Repeated query params.** The live interceptor keeps a repeated key's values as a list (`v[0] if len(v) == 1 else v`), so `?tag=a&tag=b` does not match a constant of `{"tag": "a"}` and the request is let through. The offline mirror took `v[0]` unconditionally, matched, and reported the run as intercepted. Run against `main`'s actual code:

```
OLD offline verifier (origin/main) says intercepted: True
NEW offline verifier            says intercepted: False
live interceptor params: {'tag': ['a', 'b']}
```

That is a Stage-1 pass recorded for a request the interceptor never blocked.

**2. Malformed `url_pattern`.** `server.py:455` called `re.search` unguarded inside the CDP event loop, so a bad pattern raised there and stopped interception for the remainder of the run — every later task in that run silently scoring Stage-1 zero (the edge case #258 flags). The mirror caught `re.error` and returned `False`.

`runtime-server/matching.py` is now the single copy. `server.py` imports it as a sibling (uvicorn runs from that directory per `harnesses/base/entrypoint.sh:37`), and `edgebench_judge.py` loads it by file since `runtime-server` is not a valid module path. It is kept to the standard library because the offline verifier imports it on the host, where the runtime-server's dependencies are absent.

Every failing check in `server.py`'s gate took the same action — continue the request — so the four inline branches collapse into one shared call, rather than four predicates a reader has to keep in the same order as the verifier's.

### One deviation from the issue, deliberately

The issue proposes `runtime/shared/matching.py`. That would not boot: **`harnesses/base/Dockerfile.base` copies `runtime-server/server.py` but never copies `shared/`** — only `harbor/Dockerfile` does. A module under `shared/` would be missing from every non-harbor image and `server.py` would fail its import at startup. It lives in `runtime-server/` instead, which both Dockerfiles already copy, and a test asserts the two stay in step.

### Two intended behaviour changes

Both adopt the live interceptor as the truth, since it is what actually happened during the run:

1. The offline verifier is stricter on repeated query params and now agrees with the interceptor.
2. A malformed `url_pattern` is a no-match instead of an exception, so one bad task no longer disarms interception for the rest of the run.

Happy to split (2) out if you would rather keep this PR purely mechanical.

## Corpus

- [x] v2
- [ ] v1
- [ ] both
- [ ] not applicable

Stage-1 interception is the scoring path for both corpora; ticking the default rather than claiming a V1 verification I have not run.

## Test plan

- [x] New `tests/test_stage1_matching.py` (26 tests): a 14-case fixture matrix over `url_pattern` / `method` / `body` / `params` — including batched-GraphQL list bodies, scalar bodies, empty constraints, and a forged `params` field that must not fake a match — plus both divergences and the `re.error` cases.
- [x] **Divergence #1 verified against `origin/main`'s own `_stage1_match`**, not just described: it returns `True` where the new one returns `False`, matching the transcript above.
- [x] Guards against the duplication returning: neither side may define `_const_fields_match` or call `re.search`, and `matching.py` must stay free of runtime-server dependencies.
- [x] Guards the deployment risk: both Dockerfiles must copy `matching.py` wherever they copy `server.py`. `server.py` imports it at startup, so an image with one and not the other fails to boot the runtime-server.
- [x] Existing `tests/test_edgebench_judge.py` is unchanged and green — the shared predicate is behaviour-compatible on everything that was already covered.
- [x] Full suite: 225 passed, 3 skipped. The single failure, `test_host_tasks.py::...[v1-lite]`, is a local Windows artifact (git symlinks, mode `120000`, checked out as text) and is green on CI.
- [x] `ruff check .`: 157 findings on this branch and 157 on `main` — zero introduced. `ruff format --check` clean.
- [x] `pyright src/clawbench tests`: **6 errors on this branch vs 7 on `main`**, none in the changed files.
- [x] Conflict-free against `main` and against my other open PRs (#312, #313, #314, #316, #319) via `git merge-tree`.

**What I could not verify:** there is no Docker on this machine, so the `COPY` lines and the sibling import are argued from the Dockerfiles and the `uvicorn server:app` working directory rather than observed in a built image. A maintainer running one task end-to-end would close that gap; the Dockerfile test guards the failure mode I could guard offline.

## Related issues

Fixes #301. Touches the malformed-regex edge case noted in #258.

### DIFF
--- a/src/clawbench/eval/edgebench_judge.py
+++ b/src/clawbench/eval/edgebench_judge.py
@@ -23,13 +23,37 @@ import hashlib
 import hmac
 import json
 import os
-import re
 import sys
 from pathlib import Path
 from typing import Any
-from urllib.parse import parse_qs, urlparse
 
 from clawbench.runner.judge import judge_request
+from clawbench.utils.paths import RUNTIME_ROOT
+
+
+def _load_runtime_matching():
+    """Load the Stage-1 predicate from the runtime-server directory.
+
+    It lives beside the interceptor that runs it, because runtime-server/ is
+    what gets COPYed into every task image. That directory name is not a valid
+    module path, so it is loaded by file. Sharing the one copy is the point: a
+    re-implementation here is what drifted from the live interceptor and made
+    offline verdicts disagree with real runs.
+    """
+    import importlib.util
+
+    path = RUNTIME_ROOT / "runtime-server" / "matching.py"
+    spec = importlib.util.spec_from_file_location(
+        "clawbench_runtime_matching", str(path)
+    )
+    if spec is None or spec.loader is None:  # pragma: no cover - packaging error
+        raise ImportError(f"cannot load the Stage-1 matcher from {path}")
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+_matching = _load_runtime_matching()
 
 
 def _verify_signature(intercept: dict[str, Any], secret: str) -> bool:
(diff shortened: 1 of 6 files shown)


Perry2004:
Closed for the same reason as the issue close. 

> This level of redundancy is acceptable, and I'd suggest keeping it as it is for simplicity and robustness.

PR #324, opened 2026-08-25
fix(harbor): numeric verifier metrics, lenient + strict judge rubrics, and interception correctness

rgarcia:
## Summary

The Harbor verifier now reports the same scoring rubrics as the native runner, with machine-safe outputs:

- `reward.json` carries only numeric metrics (`reward`, `intercepted`, `judge_match`, `judge_match_lenient`, `judge_match_strict`, `reward_lenient`, `reward_strict`). Full metadata — including both verdict reasons — stays in `clawbench-result.json`, so Harbor's reward aggregation never sees booleans, strings, or nulls.
- Each intercepted request is judged concurrently under two rubrics: the lenient no-explicit-contradiction rubric (the public leaderboard's primary signal) and a conservative strict rubric. The primary `reward` remains the lenient verdict; strict is reported separately as `reward_strict`. Both verdicts and reasons are retained.
- Interception evidence is timestamped: the runtime server records `intercepted_at` when it captures the request.
- The Red Cross chapter-finder task now intercepts the actual ZIP lookup API call instead of blocking the chapter-finder landing page.
- Stock Harbor agents (Claude Code and Codex) are wrapped at install time so ClawBench's existing stop signal ends them cleanly after the final request is recorded, persisting stop timing to `/data/agent-stop.json`.
- Verifier step timeout is raised from 180s to 300s to fit two judge calls.

## Why

- Harbor 0.21.0 expects numeric reward metrics; the old `reward.json` mixed booleans, strings, and nested metadata into one blob.
- Non-intercepted tasks and judge-call failures previously emitted sparse or non-numeric payloads; both paths now produce well-formed numeric metrics, with failure reasons preserved in `clawbench-result.json`.
- Stopping stock agents after interception prevents them from continuing to act or hanging the trial after the scored request has been captured.

## Relation to #319

#319 collapses `runner/judge_llm.py` onto `judge.py`'s transport in the native runner. This PR is independent of that refactor but shares its goal for the Harbor path: the verifier's lenient rubric text mirrors `judge_llm.JUDGE_SYSTEM`.

The Harbor verifier runs standalone inside each task container via the runtime-server venv where the `clawbench` package is not installed, so it cannot import the shared prompt today. If a suitable shared home for rubric strings lands, we're happy to follow up and deduplicate.

## Tests

- `uv run pytest -q` — 208 passed
- `uv run ruff check src/clawbench tests` — clean
- `uv run ruff format --check src/clawbench tests` — clean
- `uv run --frozen pyright src/clawbench tests` — 0 errors
- Coverage includes numeric metric output, dual-rubric wiring, unknown verdicts, stop-wrapper behavior, and the Red Cross interception regression.

## Reproduction (Harbor 0.21.0)

Prerequisites: Docker, `uv`, agent-provider credentials, judge credentials, and the PurelyMail variables required by ClawBench.

```bash
git clone https://github.com/kernel/ClawBench
cd ClawBench
git checkout hypeship/upstream-harbor-correctness
uv sync --frozen

uv run clawbench-harbor-adapt \
  --output-dir ./harbor-datasets/clawbench-v2-smoke \
  --task-ids v2-1134-chapter-finder-redcross \
  --overwrite

cat > .harbor-smoke.env <<EOF
PURELY_MAIL_API_KEY=$PURELY_MAIL_API_KEY
PURELY_MAIL_DOMAIN=$PURELY_MAIL_DOMAIN
OPENROUTER_API_KEY=$OPENROUTER_API_KEY
CLAWBENCH_JUDGE_BASE_URL=https://openrouter.ai/api/v1
CLAWBENCH_JUDGE_API_KEY=$OPENROUTER_API_KEY
CLAWBENCH_JUDGE_MODEL=deepseek/deepseek-v4-pro
CLAWBENCH_JUDGE_API_TYPE=openai-completions
EOF
chmod 600 .harbor-smoke.env

uvx --from harbor==0.21.0 harbor run \
  --path ./harbor-datasets/clawbench-v2-smoke \
  --agent hermes \
  --model deepseek/deepseek-v4-flash \
  --env-file .harbor-smoke.env \
  --jobs-dir ./harbor-jobs \
  --max-retries 0 \
  --delete \
  --yes

trial=$(find ./harbor-jobs -path '*/steps/run/verifier/reward.json' -print -quit)
verifier_dir=$(dirname "$trial")

cat "$verifier_dir/reward.txt"
cat "$verifier_dir/reward.json"
cat "$verifier_dir/clawbench-result.json"

rm -f .harbor-smoke.env
```

Expected: `reward.txt` equals `reward_lenient`; `reward.json` contains numeric `intercepted`, `reward_lenient`, and `reward_strict` metrics; `clawbench-result.json` retains both rubric reasons.

### DIFF
--- a/docs/harbor.md
+++ b/docs/harbor.md
@@ -45,7 +45,7 @@ uv run clawbench-harbor-adapt \
 
 ## 2. Wire up the judge
 
-Harbor's verifier calls the same judge ClawBench uses. Export the four variables once, then forward them into each run with `--ve`:
+Harbor's verifier applies both ClawBench V2 judge rubrics to every intercepted request. `reward` and `reward_lenient` use the public leaderboard's no-explicit-contradiction rubric; `reward_strict` requires the payload to demonstrate complete fulfillment. The two judge calls run concurrently against the same request and use the model configured below. Export the four variables once, then forward them into each run with `--ve`:
 
 ```bash
 export CLAWBENCH_JUDGE_BASE_URL="https://your-judge-provider.example/v1"
@@ -140,7 +140,7 @@ uvx --from harbor==0.15.0 harbor run -p ./harbor-datasets/clawbench-v2-smoke \
 
 Each converted task directory carries its own `environment/` (Chromium, the ClawBench recorder/interceptor, noVNC, runtime helper scripts), a `run/` step with `instruction.md`, the original `task.json`, the `eval-schema.json`, and a verifier under `tests/`. It deliberately contains **no ClawBench-native harness** — Harbor installs and runs whatever agent you pass to `-a` inside the task container.
 
-Scoring is the same two-stage rule as the native runner: the interceptor must catch a request matching the task schema, and the judge must agree the payload fulfills the instruction.
+Scoring uses the same two-stage rule as the native runner: the interceptor must catch a request matching the task schema, then the verifier emits both the public lenient reward and the conservative strict reward. Harbor uses the lenient result as the primary `reward` metric and retains both verdicts and reasons in `clawbench-result.json`.
 
 ## Troubleshooting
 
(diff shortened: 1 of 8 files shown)


Perry2004 (approved):
Thanks for the contribution. That's a good enhancement on the adaptor. I'll merge to main.

PR #328, opened 2026-08-31
docs(scoring): make eval/scoring.md the single source

### LINKED ISSUE #293: docs: eval/scoring.md and docs/scoring.md are duplicate copies that have started to drift (one points at the deprecated leaderboard)
`eval/scoring.md` and `docs/scoring.md` are both 190 lines and ~identical — but they have already started to drift, and the drift is in the links that matter:

| | `eval/scoring.md` | `docs/scoring.md` |
|---|---|---|
| Live leaderboard | `huggingface.co/spaces/NAIL-Group/clawbench-leaderboard` (superseded — kept only as a historical mirror per the 2026.05.12 news entry) | `huggingface.co/spaces/TIGER-Lab/ClawBench` (canonical) |
| HF data card | `datasets/NAIL-Group/ClawBench` | `spaces/TIGER-Lab/ClawBench` (points at a Space, not a dataset) |

README links **both**: line 190 sends readers to `docs/scoring.md`, line 268 sends them to `eval/scoring.md`. zh-CN links `eval/scoring.md` twice. So a reader following "the scoring formula" gets one of two documents depending on which link they clicked, and one of them steers them to a deprecated leaderboard.

**Fix:** pick one canonical location (`docs/scoring.md` matches where the rest of the prose docs now live after #283), make the other a one-line stub pointing at it, and update the four inbound links. Also fix the `spaces/` vs `datasets/` mixup on the data-card row while touching it.

Related: this class of bug is what #291's markdown link checker guards against structurally, but it can't catch two files that both exist and disagree.

vaibhavdabas16:
## What does this PR do?

Fixes #293.

`docs/scoring.md` and `eval/scoring.md` were 190-line copies of each other, drifted apart on the four links that decide which numbers a reader tries to reproduce. **Neither copy was right on all four**, which is the part worth flagging before anything else:

| | correct (`README.md`, `docs/news.md`) | `docs/scoring.md` | `eval/scoring.md` |
|---|---|---|---|
| Live leaderboard | `spaces/TIGER-Lab/ClawBench` | correct | **stale** — `spaces/NAIL-Group/clawbench-leaderboard` |
| HF data card | `datasets/NAIL-Group/ClawBench` | **wrong** — points at a *Space*, not a dataset | correct |
| V1 trace | `datasets/NAIL-Group/ClawBenchV1Trace` | **label/URL disagree** — label says TIGER-Lab, URL goes to NAIL-Group | correct |
| V2 trace | `datasets/TIGER-Lab/ClawBenchV2Trace` | correct | **stale** — NAIL-Group |

So deduping toward *either* file verbatim would have shipped two wrong links. The "correct" column is what `README.md:12`, `:110`, `:13`, `:147`, `:425-427` and `docs/news.md:15` record.

### Which file is canonical

`eval/scoring.md`. It sits beside `eval/rescore.py`, which implements it, and `README.md`, `docs/README.zh-CN.md` and `docs/harbor.md` already link there — `README.md:433` names it as *the* scoring formula.

**This differs from the issue**, which proposed `docs/scoring.md` on the grounds that prose docs moved to `docs/` in #283. That's a fair argument and I'd switch on request; the reason I didn't is that it would mean redirecting the three inbound links that currently point at `eval/scoring.md`, including the README's own. Happy to flip it if you'd rather follow #283's layout — the link corrections below are the same either way.

`docs/scoring.md` becomes a short pointer so existing links, including any outside this repo, keep resolving. The two in-repo references that treated it as the spec (`docs/v1-vs-v2.md`'s provenance line, `docs/news.md`'s "Details" link) now go straight to the canonical file.

### The link corrections

Removing a duplicate means choosing values, not just a filename. The survivor takes the README-corroborated URL on every row: its leaderboard and V2 trace links are corrected to TIGER-Lab; its already-correct dataset card and V1 trace links stay NAIL-Group.

### What this deliberately does not do

It does not settle the TIGER-Lab/NAIL-Group split in general. That runs through `docs/news.md`, `docs/README.zh-CN.md`, `docs/v1-vs-v2.md` and `docs/huggingface-space-metadata.md` as well, and reconciling those needs maintainer knowledge of which repos are actually live. All this pins is that the scoring spec agrees with the README.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

<sub>Documentation only; no task data and no runtime code touched.</sub>

## Test plan

- [x] New `tests/test_docs_scoring_single_source.py` (13 tests): the `docs/` copy stays a pointer rather than a rival spec, each spec section appears in exactly one file, nothing still references the old path as the spec, the pointer's relative links and anchors resolve, and each of the four URLs matches `README.md`.
- [x] **Controls, because a test written after the fix proves nothing.** Five of the original eight assertions fail against the pre-change tree. For the four new link assertions, holding the tests constant and reverting only `eval/scoring.md` fails 3 of 5 — the two corrected URLs plus the superseded-Space guard.
- [x] The whole-URL matcher is deliberate: a plain substring check for `.../ClawBench` is satisfied by `.../ClawBenchV1Trace`, so the dataset-card assertion would have passed vacuously.
- [x] Full suite: 223 passed, 4 skipped. The 4 deselected are `test_host_tasks.py::test_checked_task_json_files_parse_and_validate`, which fails only on this Windows checkout — the task files are git symlinks (mode `120000`) checked out as text — and is green on CI.
- [x] `ruff check`, `ruff format --check`, and `pyright --pythonplatform Linux` (matching the ubuntu runner) all clean.

## Related issues

Fixes #293. Related: #291's markdown link checker guards this class of bug structurally, but cannot catch two files that both exist and disagree — which is what happened here.


Perry2004:
I've pushed a bit cleanup to it, the rest looks good. 
There seem to be a GitHub outage that the PRs not updating, I'll approve & merge after it's updated. 

Thanks for the contribution.

Issue #160, opened 2026-05-13
test(resilience): verify resume works — auto-restart on abort + alerting


reacher-z:
## Motivation

Long V1/V2 runs take **8–20 hours per (model × harness)** in our recent runs (per `wall_hours` in [`leaderboard/results.csv`](https://huggingface.co/datasets/TIGER-Lab/ClawBench/blob/main/leaderboard/results.csv)). Today's failure modes that show up in the wild:

- **Container OOM** — Chromium balloons mid-run, kernel kills it, batch aborts at task N/130
- **Network blip** — Anthropic/OpenAI/OpenRouter 502 for 30s, agent times out, batch reports task as failed (but really it was an infra issue, not the agent's)
- **Site captcha mid-run** — agent gets through 60 tasks then hits a captcha on task 61, blocked for the rest
- **Host VM evicted** — Vultr / cloud reboot in the middle of a multi-hour batch
- **Queue cap** — `claude-opus-4-7` hit our queue cap at task 75 (partial), `gpt-5.5` at 81 (partial) — should those have auto-resumed when the cap reset?

Right now the script just **stops**, leaving a `(model, harness, corpus)` cell partial on the leaderboard (\"*partial*\" tag on 2/6 V2 rows already). Every partial is a small operator-pain point: someone has to notice, manually restart, and re-merge results.

This issue asks: **prove resume works end-to-end**, then add auto-restart + alerting where it doesn't.

### Test 1 — Manual resume mid-batch

```bash
# Start a V2 hermes run on a cheap model
clawbench run --corpus v2 --model deepseek-v4-flash --harness hermes &
PID=$!
sleep 600          # 10 min in — should be ~5–10 tasks deep
kill $PID          # simulate abort
clawbench run --corpus v2 --model deepseek-v4-flash --harness hermes  # re-invoke
```

Expected: the second invocation **picks up where the first left off** (doesn't re-run tasks 1–10). Verify by:
- Counting trace dirs under `claw-output/v2-results/deepseek-v4-flash_hermes_v2/` — should be ≥10 after the kill, then continue from there
- `results.csv` post-completion should reflect exactly 130 tasks, not 140

### Test 2 — Mid-task abort

Kill while task #42 is mid-execution (agent has been clicking around for 3 minutes). Re-invoke. Does the resume:
- (a) Skip task #42 (assume done) — **wrong**, lost data
- (b) Re-run task #42 from scratch — **expected**
[...]

PR #329, opened 2026-08-31
fix(batch): resume from run-meta.json, not from log-file existence

### LINKED ISSUE #297: batch --resume skips failed jobs (log-file heuristic) and overwrites batch-summary.json with 'skipped' rows
`src/clawbench/runner/batch.py:651-655`:

```python
for job in jobs:
    safe_model = re.sub(r"[/:]+", "--", job.model)
    if (log_dir / f"{job.case_name}-{safe_model}.log").exists():
        job.status = "skipped"
```

The log file is written unconditionally when a job *starts* (`batch.py:320`), so `--resume` skips every job that was ever attempted — including the ones that **failed or errored**. That inverts the flag's main use case: after a network blip or a container OOM kills 20 of 130 tasks, `--resume` skips exactly those 20 and reports the batch complete.

Second effect: after resume, `write_summary_json` (`batch.py:550-579`, called at `:754`) rewrites `batch-summary.json` from the in-memory job list, so every previously-completed job is recorded as `status: "skipped", duration_seconds: 0`. The original tallies are destroyed in the artifact that downstream stats and the HF upload consume.

**Ask:**
1. On resume, decide completion from each run's `run-meta.json` (authoritative outcome), not from log-file existence.
2. Add `--retry-failed` to re-run jobs whose recorded outcome was failure/infra error.
3. Merge prior job statuses into the rewritten summary rather than overwriting them with `skipped`.

Related: #160 (resume/resilience testing) asks for exactly this path to be verified end-to-end; there is currently no test covering resume (`tests/test_batch_and_tui_helpers.py` only passes `resume=None`).

vaibhavdabas16:
## What does this PR do?

Fixes #297 — all three asks.

`--resume` decided a job was finished if `batch-logs/<case>-<model>.log` existed. That log is opened when a job **starts** (`batch.py:320`), so resume skipped every job that was ever *attempted* — including the ones killed mid-run. That inverts the flag: after a network blip or an OOM takes out 20 of 130 tasks, `--resume` skips exactly those 20 and reports the batch complete.

The second effect destroyed data rather than just omitting it. `write_summary_json` rewrites `batch-summary.json` from the in-memory job list, and every carried-over job was written as `status: "skipped", duration_seconds: 0`. A resumed batch reported none of the results it already had, in the artifact the per-run stats and the HuggingFace upload consume.

### Ask 1 — decide completion from `run-meta.json`

`run.py` writes `run-meta.json` for any run that got far enough to have an outcome, failures included — the invariant #303, #302 and #314 each protect. `load_recorded_runs()` indexes those by `(test_case, model)`, read from the metadata rather than the path, so:

- a model name containing `/` matches its job even though the run directory sanitises it to `--`;
- the newest run wins when repeated resumes have left several directories for one `case × model` pair;
- a **truncated** `run-meta.json` counts as an unknown outcome and the job runs again, rather than having a result invented for it. #312/#325 established that this file can be truncated.

A job with no recorded outcome stays pending and runs again. That is the whole fix for ask 1: "has a log" becomes "has a result".

### Ask 2 — `--retry-failed`

Re-runs jobs whose recorded outcome was an infra-class failure, using the scorer's own `NON_MODEL_FAILURE_CATEGORIES` (`infra_failure`, `api_or_credit`, `task_data`, `build_instruction`). Those are already the categories excluded from adjusted scoring, so such a run carries no signal about the model and a re-run is the only way to get one.

**One judgment call worth your eye:** genuine model failures (`model_not_intercepted`) are *kept*, not retried — the model was asked and it failed, so a re-run buys the same answer at full agent cost. The issue says "failure/infra error", which could be read either way. Say the word and I'll widen it to retry those too, or split it into two flags.

### Ask 3 — merge prior statuses into the summary

A carried-over job keeps its recorded result — `passed` / `failed` / `error` — with its real duration, so the rewritten `batch-summary.json` preserves the original tallies instead of zeroing them.

Jobs are marked `resumed`, which does two things: the scheduler still skips executing them (it previously keyed off `status == "skipped"`, which is no longer what a carried-over job looks like), and a reader of `batch-summary.json` can tell a carried-over row from one this invocation produced. That adds one key per job to the summary — flagging it since that file is consumed downstream, though it is additive and existing keys are unchanged.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

<sub>Host-side batch driver; no task data involved.</sub>

## Test plan

- [x] New `tests/test_batch_resume.py`, 17 tests. This path had **no** coverage — every existing test passes `resume=None`, which #297 notes and #160 asks to fix.
- [x] **Control, because tests written after a fix prove nothing.** I swapped the log-existence heuristic back into `apply_resume`, held the tests constant, and re-ran: **8 of the 17 fail**, including the two that matter most — a started-but-unfinished job being skipped, and the summary losing its tallies.
- [x] Covered: a job with a start log but no metadata runs again; a finished job does not; carried-over status and duration survive into `batch-summary.json` (`totals` stay `passed: 1, failed: 1` rather than collapsing to `skipped: 2`); the outcome→status mapping; missing/unparsable `duration_seconds`; `--retry-failed` on and off; newest-run-wins; truncated metadata; a `/` in the model name; an empty or absent output directory.
- [x] Full suite: 227 passed, 4 skipped. The 4 deselected are `test_host_tasks.py::test_checked_task_json_files_parse_and_validate`, which fails only on this Windows checkout — the task files are git symlinks (mode `120000`) checked out as text — and is green on CI.
- [x] `ruff check` reports the same 8 pre-existing findings on `batch.py` as `main` does — no new ones. `ruff format --check` clean. `pyright --pythonplatform Linux` (matching the ubuntu runner) 0 errors.
- [x] `batch.py` must stay importable without a container engine (`test_batch_stays_importable_without_a_container_engine`). The new import is `run_support.results`, which does not pull in `run_support.config`; verified, and that test still passes.
- [x] Merge-clean against `main` and against all five of my other open PRs — including #316, which touches the same file.

**Not verified:** no live batch run — I don't have Docker on this machine. The resume decision logic is exercised against real-shaped `run-meta.json` fixtures on disk, not against an interrupted 130-task sweep.

## Related issues

Fixes #297. Related: #160 (resume/resilience testing) asks for exactly this path to be verified end-to-end; this adds the unit-level half.

<sub>`docs/cli.md`'s `--resume` row described skipping "finished runs", which was the intent but not the behaviour — updated to match what the code now does, plus a row for `--retry-failed`.</sub>

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -73,7 +73,8 @@ Execution:
 | --- | --- | --- |
 | `--max-concurrent <n>` | 2 local, 1 Kernel/Browserbase | Parallel jobs |
 | `--stagger-delay <s>` | 15 | Minimum seconds between consecutive container starts (rolling start) |
-| `--resume <dir>` | — | Reuse a previous batch's output directory and skip finished runs |
+| `--resume <dir>` | — | Reuse a previous batch's output directory and skip jobs whose `run-meta.json` records an outcome. Jobs that started but never finished are run again |
+| `--retry-failed` | off | With `--resume`, also re-run jobs whose recorded outcome was an infra-class failure (the categories excluded from adjusted scoring). Genuine model failures are kept |
 | `--dry-run` | off | Print the job matrix without running anything |
 | `--output-dir <path>` | `test-output` | Base output directory |
 
--- a/src/clawbench/runner/batch.py
+++ b/src/clawbench/runner/batch.py
@@ -18,6 +18,7 @@ from typing import Protocol
 import yaml
 
 from clawbench.runner.run_support.config import engine
+from clawbench.runner.run_support.results import NON_MODEL_FAILURE_CATEGORIES
 from clawbench.utils.paths import ASSET_ROOT, WORKSPACE_ROOT, ensure_workspace_templates
 from clawbench.utils.timeouts import (
     BATCH_JOB_GRACE_S,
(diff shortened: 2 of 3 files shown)


vaibhavdabas16:
Rebased onto `main` at `9dd9d44`. One behavioural follow-up in `a5f219b`: main now has a `judge_inconclusive` batch status (#6a7de9c gave a judge outage exit 3), so `recorded_outcome()` now mirrors the live exit-code map — it reads run-meta's `pass` rather than `intercepted`, and an intercepted run whose `judge_match` is `null` resumes as `judge_inconclusive` instead of being upgraded to a pass. Tests cover both new rows; CI green.

PR #311, opened 2026-08-19
feat(browser-runtime): add native Kernel support

rgarcia:
## What does this PR do?

Adds Kernel as a managed browser runtime:

```bash
uv run clawbench-run test-cases/v2/<case> <model> \
  --browser-runtime kernel \
  --browser-runtime-options '{"stealth":true,"region":"us-east"}'
```

The integration creates a Kernel browser, connects the existing ClawBench runtime and harness over CDP, captures the standard action/request/screenshot/message artifacts, downloads the Kernel replay as `data/recording.mp4`, and deletes the browser after the run.

No new package dependency is required; the provider uses Kernel's REST API through the standard library.

## Lifecycle

The new `KernelRuntimeProvider` handles:

1. Browser creation using `KERNEL_API_KEY` and optional `KERNEL_BASE_URL`.
2. Runtime options for `stealth`, `region`, `proxy`, and `tags`.
3. Replay creation before the harness connects.
4. Replay stop, processing polling, documented HTTP 202 retry handling, and MP4 download.
5. Browser deletion on success, failure, interruption, or replay errors.

Replay finalization uses a shared five-minute deadline. Missing or invalid recordings produce `browser_runtime_artifact_failed` rather than silently marking the run successful.

## Managed-runtime behavior

This also generalizes behavior that was previously specific to Browserbase:

- managed runtimes default to batch concurrency 1;
- batch runs hide credential-bearing viewer URLs;
- recording behavior is selected through provider capabilities;
- `claude-code-chrome-extension` is rejected for managed runtimes because it requires its own browser process inside the container.

Local, remote-CDP, Steel, and Browserbase behavior remains unchanged.

## User-facing changes

- Adds `kernel` to CLI and TUI runtime selection.
- Documents configuration and runtime options.
- Adds provider-neutral managed-runtime messaging.
- Updates the English and Chinese documentation and changelog.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

## Test plan

- [x] `uv run ruff format --check .`
- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest -q` — 203 passed
- [x] Real Kernel + browser-use run completed with interception and all five artifact layers.
- [x] Kernel replay downloaded as a playable MP4 and the browser was deleted.
- [x] 20-task V2 Hermes batch at concurrency 4 completed with zero infrastructure failures; all 20 replays downloaded and all 20 browsers were deleted.
[...]

PR #330, opened 2026-08-31
feat(harbor): add Kernel browser control arm

rgarcia:
### Summary

Extend `clawbench-harbor-adapt` with a Kernel-backed browser runtime so Harbor agents can run ClawBench tasks against one managed remote browser per trial.

- add `--browser-runtime kernel` with validated runtime options
- create one headful stealth Kernel browser during task setup
- connect the agent's pinned Playwright MCP server through a credential-free local CDP bridge
- finalize and download provider replay recordings during verification
- delete the browser on success and setup failure, then verify deletion
- preserve local Chromium as the default Harbor runtime

### Why

The native ClawBench runner supports Kernel browsers after #311, while the Harbor adapter currently supports only container-local Chromium. This adds the corresponding Harbor control arm so the same task corpus can be compared across local and managed browser runtimes without changing task semantics or scoring.

The implementation is rebased on the Harbor correctness changes merged in #324. This PR contains only the remaining remote-browser Harbor integration.

### Design

Each generated Kernel task:

1. creates exactly one browser and provider replay during setup
2. stores the provider CDP endpoint in a root-only runtime file
3. exposes a local CDP bridge to Playwright MCP
4. records credential-free session and lifecycle metadata under `/my-info/`
5. finalizes the replay, writes `recording.mp4`, deletes the browser, and verifies deletion

The existing local Chromium path remains the default and does not require Kernel credentials.

### Validation

- `uv run pytest -q` — 232 passed
- `uv run ruff check src/clawbench tests`
- `uv run ruff format --check src/clawbench tests`
- `uv run --frozen pyright src/clawbench tests` — 0 errors
- fork CI passes on Ubuntu, macOS, and Windows

### DIFF
--- a/src/clawbench/eval/harbor_adapter.py
+++ b/src/clawbench/eval/harbor_adapter.py
@@ -11,11 +11,24 @@ import sys
 from pathlib import Path
 from typing import Any
 
+from clawbench.runner.run_support.browser_runtime.providers import (
+    BrowserRuntimeError,
+    _parse_options,
+)
 from clawbench.runner.run_support.task import build_instruction, validate_task_data
 from clawbench.utils.paths import RUNTIME_ROOT, asset_path
 
 DEFAULT_CASES_DIR = asset_path("test-cases", "v2")
 STEP_NAME = "run"
+HARBOR_BROWSER_RUNTIMES = ("local", "kernel")
+# Remote browser runtimes connect through the runtime server's local,
+# credential-free CDP bridge instead of a container-local Chromium.
+REMOTE_BRIDGE_CDP_URL = "http://127.0.0.1:7878"
+LOCAL_CDP_URL = "http://127.0.0.1:9223"
+# Pinned Playwright MCP package Harbor's stock Claude Code and Codex agents
+# use to drive the ClawBench browser through the CDP bridge.
+PLAYWRIGHT_MCP_PACKAGE = "@playwright/mcp"
+PLAYWRIGHT_MCP_VERSION = "0.0.79"
 
 
 def sanitize_task_name(raw: str) -> str:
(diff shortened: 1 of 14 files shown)


src/clawbench/runtime/harbor/start-runtime.sh
```diff
@@ -8,16 +8,32 @@ if [ -f /tmp/clawbench-run/runtime.pid ] && kill -0 "$(cat /tmp/clawbench-run/ru
...
+if [ -n "${CLAWBENCH_BROWSER_CDP_URL_FILE:-}" ] && [ -f "${CLAWBENCH_BROWSER_CDP_URL_FILE}" ]; then
+  REMOTE_MODE=true
+fi
 
 cd /app/src/runtime-server
 uv run --no-sync uvicorn server:app --host 0.0.0.0 --port 7878 >/tmp/clawbench-run/runtime-server.log 2>&1 &
 echo "$!" > /tmp/clawbench-run/runtime-server.pid
 sleep 1
 
+if [ "$REMOTE_MODE" = true ]; then
```
Perry2004, line 31: This should be moved to before runtime-server start since the server starts the ffmpeg in local mode, which depends on x11y display.

src/clawbench/eval/harbor_adapter.py
```diff
@@ -108,18 +131,71 @@ def copy_environment(env_dir: Path) -> None:
...
-        "- CDP endpoint: http://127.0.0.1:9223\n"
+        "- Use the existing browser session exposed by Chrome DevTools Protocol.\n"
+        f"- CDP endpoint: {cdp_url}\n"
         "- CDP environment variables are also set for the agent process: "
         "CLAWBENCH_CDP_URL, BROWSER_CDP_URL, CDP_URL, CHROME_CDP_URL, and PLAYWRIGHT_CDP_URL.\n"
-        "- noVNC viewer, if needed: http://127.0.0.1:6080/vnc.html\n"
-        "- Do not launch a separate browser. Complete the task through the existing browser session.\n"
-        "---\n"
+    )
+    if browser_runtime == "local":
```
Perry2004: Please do not change the prompt to the agent. These additional info will have potential impact on agent performance.

Perry2004 (changes requested):

rgarcia:
thanks for the review @Perry2004! Just addressed your comments

Perry2004 (approved):
LGTM🫡
Thanks for the contribution.

PR #333, opened 2026-08-31
fix(judge): give a judge outage its own outcome, not JUDGE MISMATCH

### LINKED ISSUE #299: Judge outages are scored as agent failures (match=None → exit 1 → 'failed')
When the judge provider is down, we record the model as having *failed the task*.

- `src/clawbench/runner/judge.py:247-253` returns `match=None` on `judge_call_failed` (HTTP 402/timeout after retries).
- `src/clawbench/runner/run.py:657-664` marks `pass` only when match is `True`; `run.py:743-754` then exits 1.
- `src/clawbench/runner/batch.py:324-327` counts exit 1 as `failed`.

So a judge outage is indistinguishable in `batch-summary.json` from genuine agent failure. This is not hypothetical — a judge account hitting HTTP 402 mid-sweep has already produced whole batches recorded as failures that needed a manual rescore pass to recover.

**Ask:**
1. Give `match=None` a distinct outcome (e.g. exit code 3 / `status: "judge_inconclusive"`), separate from agent failure.
2. Count it separately in `batch-summary.json` totals so a reader can see "N runs unjudged" instead of silently deflated reward.
3. Have `--resume` / `clawbench-rescore` target exactly those runs for re-judging.

This pairs with #243 (always report Stage-1 and Stage-2 separately): both are about not letting stage-2 problems silently corrupt the headline number.

AmirF194:
## What does this PR do?

Fixes #299, all three asks.

`judge.py` already returns `match=None` when the judge never renders a verdict (provider outage, retries exhausted, or an unparseable reply), distinct from `match=False` (a genuine mismatch). But downstream, `run.py`'s final-status block treats both the same way, sharing exit 1, and `batch.py`'s exit-code map turns exit 1 into `job.status = "failed"`. So a judge-provider outage was indistinguishable from an agent failure in `batch-summary.json`.

### Ask 1: give match=None a distinct outcome

`run.py` now exits 3 instead of 1 when `judge_result.get("match") is None`, keeping it separate from a genuine `JUDGE MISMATCH` (which still exits 1).

### Ask 2: count it separately in batch-summary.json

`batch.py` maps exit 3 to `job.status = "judge_inconclusive"`, its own bucket alongside `passed`/`failed`/`error`/`skipped`, carried into `batch-summary.json`'s `totals` and the printed batch summary.

### Ask 3: retarget those runs for re-judging

`clawbench-rescore` now retries a cached verdict whose `match` is `null` even without `--force`, since that cached file never held a scored result. `--resume`'s own skip logic in `batch.py` is untouched here; it currently keys off batch-log existence rather than the recorded outcome, and reworking that is a larger, separate change.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

<sub>Host-side runner change; no task data involved.</sub>

## Test plan

- [x] `tests/test_run_judge_stage.py::test_judge_failure_after_the_run_still_writes_run_meta`, updated: a `judge_setup_failed` outcome now exits 3, not 1. New `test_judge_mismatch_keeps_exit_code_1_distinct_from_inconclusive` locks in that a real `match=False` still exits 1.
- [x] `tests/test_batch_judge_inconclusive.py`, new: `run_job()` against a mocked subprocess maps exit 0/1/3/other to `passed`/`failed`/`judge_inconclusive`/`error`; `write_summary_json()` counts `judge_inconclusive` separately from `failed` in `batch-summary.json`.
- [x] `tests/test_rescore_retarget.py`, new: `rescore_one()` retries a cached `match: null` verdict without `--force`, and still trusts a cached scored verdict (`match: true`/`false`) without retrying it.
- [x] Full suite in a clean `python:3.11-slim` container via `uv run --frozen pytest`: 226 passed (218 on `main`, +8 from this PR). All new/changed tests fail on `main` and pass on this branch, confirmed both ways in the same container.
- [x] `ruff check .`, `ruff format --check .`, `pyright src/clawbench tests`, `uv build` + `twine check dist/*`, and the three `--help` console-script checks from `static-check.yml`: all clean. `validate-docs.yml`'s link checker (triggered by the `CHANGELOG.md` edit): clean.

**Not verified:** no live judge-provider outage was reproduced end-to-end; the fix is proven through `judge.py`'s existing `judge_call_failed`/`unsupported_api_type` fallback plus the tests above, not a real API timeout.

## Related issues

Fixes #299.


AmirF194:
#338 lands essentially the same fix: a distinct status for a judge that never rendered a verdict, kept out of the failed bucket, touching the same three files (run.py, batch.py, rescore.py). It also adds a --only-unjudged rescore flag mine doesn't have. I'll close this one if you'd rather take that version instead.

Perry2004:
Merged with main and pushed a minor fix.

Perry2004 (approved):
LGTM after the fix🫡
Previously it is skipping those with `match: null`.

AmirF194:
Thanks for sorting the merge conflict and catching the match: null skip, appreciate the fix.

Issue #331, opened 2026-08-31
Harbor: strict harbor/terminal-bench format compliance + e2b sandbox support


reacher-z:
## Goal

Make ClawBench a **first-class Harbor benchmark**, strictly following the conventions used by [harbor-framework/harbor](https://github.com/harbor-framework/harbor) and [harbor-framework/terminal-bench](https://github.com/harbor-framework/terminal-bench), including running on Harbor's remote sandboxes (**`-e e2b`**) — not just local Docker.

Today `clawbench-harbor-adapt` (added in v0.7.0, #232) produces Harbor tasks that load fine on Harbor 0.22, but we are not yet aligned with the ecosystem end-to-end.

### 1. Strict task format (terminal-bench parity)
- [ ] Canonical layout per task: `task.toml` (`[task] name = "<org>/<name>"`, authors, keywords), `instruction.md`, `environment/` (or `[environment].docker_image`), `tests/test.sh` → numeric `/logs/verifier/reward.txt` + `reward.json`, `solution/solve.sh` (documented no-op — live sites have no scripted oracle)
- [ ] `dataset.toml` manifest with per-task digests (`harbor add --scan` / Packager content hash)
- [ ] Every task loads with Harbor `TaskConfig` in CI; committed dataset regenerates deterministically (`scripts/harbor/regenerate.sh`, `validate-harbor.yml` — on branch `feat/harbor-hub-registry`)
- [ ] Numeric-only `reward.json` metrics (depends on #324 — should merge first)

### 2. Registry + Hub (like terminal-bench)
- [ ] Root `registry.json` so `harbor run --repo TIGER-AI-Lab/ClawBench -d clawbench-v2` works
- [ ] Publish `tiger-ai-lab/clawbench-v2` to the Harbor Hub (`harbor publish --public`), tags per release
- [ ] Prebuilt runtime image on GHCR (`clawbench-harbor-runtime:<version>`) so tasks ship no build context

### 3. e2b (and other remote sandbox) support
- [ ] `harbor run ... -e e2b` completes a smoke task end-to-end (also verify `-e daytona`/`modal` are not broken)
- [ ] No dependency on local podman/docker specifics: image pullable from a public registry, no bind mounts outside the task dir, healthcheck-driven readiness (Chromium + CDP + interceptor)
- [ ] Judge + PurelyMail credentials passed via `--ve`/`--env-file` only; document resource floor (1 CPU / 2 GB per trial)
[...]

PR #341, opened 2026-09-02
docs(harbor): pin harbor==0.22.0 and make the pin checkable

### LINKED ISSUE #294: chore(harbor): docs pin harbor==0.15.0 but upstream is 0.21.0 with a changed dataset interface
Every Harbor invocation in our docs pins `harbor==0.15.0`:

- `README.md:470`, `:492`, `:507`, `:834`
- `src/README.md:29`
- (and the adapter docs generated from these)

Upstream is now **0.21.0** (harbor-framework/harbor, PyPI `harbor`). Between those versions the dataset interface changed: 0.15's local-path form `harbor run -p ./harbor-datasets/clawbench-v2` has given way to hub-published datasets addressed as `harbor run -d org/dataset@version`, and the hub publish flow (`harbor auth login` → `harbor dataset init` → `harbor add` → `harbor publish --public`) is now the documented path.

**Impact:**
1. Users following our docs install a six-versions-stale Harbor; anyone who already has current Harbor finds our commands don't match their CLI.
2. It blocks the more valuable thing: publishing ClawBench to the Harbor hub as `tiger-ai-lab/clawbench`, which makes the benchmark runnable by anyone with `harbor run -d tiger-ai-lab/clawbench@v1` and gets hub-side leaderboard/analytics for free. That publish flow requires ≥0.16-era CLI.

**Ask:**
1. Smoke-test `clawbench-harbor-adapt` output against Harbor 0.21 and note the supported range in the docs (or bump the pin).
2. Decide whether to publish the V2 dataset to the Harbor hub (separate issue if yes).

vaibhavdabas16:
Closes #294 (ask 1). Ask 2 — whether to publish the V2 dataset to the Harbor hub — is a maintainer decision and stays in #331.

Eight Harbor invocations across `docs/harbor.md` and `src/README.md` pinned `harbor==0.15.0`. Upstream is **0.22.0**. Readers following our docs installed a six-releases-stale Harbor; readers who already had a current one found our pin fighting their install.

Bumping the number is the easy half. The issue's actual ask is *"smoke-test `clawbench-harbor-adapt` output against Harbor 0.22 and note the supported range"* — a pin nobody has exercised is how we got here — so this PR does the check and leaves behind a way to repeat it.

## Verified against harbor 0.22.0

- **All 129** tasks from `clawbench-harbor-adapt` construct as `harbor.models.task.task.Task` with no errors, and `Task.is_valid_dir` accepts every generated directory.
- Every flag these docs use still exists on `harbor run`: `-p`, `-a`, `-m`, `--ak`, `--env-file`, `--ve`, `--jobs-dir`, `-n`, `-k`, `-r`, `--timeout-multiplier` and the four per-phase multipliers.

**Not** verified, and `docs/harbor.md` says so: no end-to-end `harbor run` against live sites. That needs Docker, a browser container per trial and judge credentials. Loading proves the format is accepted; it does not prove a sweep passes.

## One correction to the issue

> 0.15's local-path form `harbor run -p ./harbor-datasets/clawbench-v2` has given way to hub-published datasets addressed as `harbor run -d org/dataset@version`

Both exist in 0.22.0:

```
| --path      -p  <path>  Path to a local task or dataset directory
| --dataset   -d  <str>   Dataset name@version (e.g., 'dataset@1.0')
```

`-d` is an **addition** for hub datasets, not a replacement. So the local-path commands in these docs were never broken by the version drift — only the pin was. Worth stating plainly, because "our documented commands don't work on current Harbor" and "our pin is stale" call for different amounts of work, and it is the second one.

## Making the next bump checkable

`tests/test_harbor_version_compatibility.py` re-runs the task-loading half of the check. Harbor is not a ClawBench dependency, so it skips unless Harbor is importable:

```bash
uv run --with harbor==<new-version> pytest tests/test_harbor_version_compatibility.py
```

It converts the **real V2 corpus** through `harbor_adapter.main()` rather than a hand-built fixture — a synthetic task would only prove the fixture loads. What has to hold is that the corpus we tell people to convert produces a dataset Harbor accepts.

A second test records the installed Harbor version and skips with a clear message when it differs from the one the docs claim, so a mismatch is visible in test output rather than silently passing under a version nobody documented.

## Verification

- Default suite: `213 passed, 5 skipped` — the new file skips, no change to `main`'s count. (`test_host_tasks.py::…[v1-lite]` fails on `main` too: `v1-lite` symlinks check out as text on Windows. Unrelated.)
- With Harbor present: `2 passed in 5.44s` against `harbor==0.22.0`.

## Merge notes

Docs and one new test file only; no source changes. `CHANGELOG.md` conflicts textually with my other open PRs (#339, #340); trivial either way.


Perry2004 (approved):
Simplified the `harbor.md` a bit. The rest LGTM.

PR #344, opened 2026-09-09
feat(browser-runtime): implement the Steel browser runtime

### LINKED ISSUE #105: FEAT: support steel browser
https://github.com/steel-dev/steel-browser

vaibhavdabas16:
Closes #105. That was the last open sub-issue of #103 (support more browser runtimes).

## What

`--browser-runtime steel` has been reserved since the runtime abstraction landed, but selecting it raised `steel browser runtime is reserved but not implemented yet`. This implements it.

Steel Cloud and the self-hosted [steel-dev/steel-browser](https://github.com/steel-dev/steel-browser) linked in #105 expose the same `/v1/sessions` API, so one provider covers both:

- **Steel Cloud** — `STEEL_API_KEY` in `.env.local`, sent as the `steel-api-key` header.
- **Self-hosted** — `STEEL_BASE_URL=http://localhost:3000`. No key is required, so the key is only mandatory when the endpoint is the cloud one.

## Details

- Sessions are created with ClawBench's 1920x1080 viewport and the task time limit plus 120s of headroom. Steel expresses `timeout` in **milliseconds**, so the value is converted and bounded to `[60_000, 86_400_000]`.
- `dimensions` and `timeout` are set by ClawBench and rejected as user options; the rest of the allowlist is `blockAds`, `solveCaptcha`, `useProxy`, `proxyUrl`, `region`, `userAgent`, `stealthConfig`, `sessionContext`, `extensionIds`, each type-checked at construction the way the Browserbase and Kernel providers do.
- Steel serves its rrweb replay from the session viewer rather than as a downloadable file, so there is no `recording.mp4` to fetch. The runtime reports `recording_mode: "provider"` with the viewer URL as `recording_url` — the same shape Browserbase already uses, which `batch.py` summary rendering already understands. `finalize()` is therefore a no-op; `cleanup()` releases the session and treats 404/409 as `already_closed`.
- `steel` joins `MANAGED_BROWSER_RUNTIMES`, so batch runs default to `--max-concurrent 1`, pass `--hide-browser-viewer`, and refuse the `claude-code-chrome-extension` harness — consistent with the other managed runtimes.
- The TUI runtime picker offers Steel and prints a session-viewer tip.
- Browserbase and Kernel each carried an identical inline block scrubbing the API key out of `URLError` reasons. Steel needed a third, so that is now a shared `_scrub_secret` helper.

## Testing

`tests/test_browser_runtime.py` replaces the "reserved" assertion with 11 cases covering the create/release round trip, payload and timeout bounds, cloud vs self-hosted key requirements, option validation, release-on-bad-response, 404 cleanup, and that neither HTTP nor network errors leak the key into an error message or into `run-meta.json`.

Full suite passes locally (`286 passed, 10 skipped`).

**Not verified against a live Steel deployment** — I don't have Steel credentials. Every request/response shape is exercised against fakes; a maintainer with an account or a local `steel-browser` should smoke-test one task before merge.

### DIFF
--- a/src/clawbench/runner/run_support/browser_runtime/providers.py
+++ b/src/clawbench/runner/run_support/browser_runtime/providers.py
@@ -31,6 +31,18 @@ _KERNEL_ALLOWED_OPTIONS = {
     "stealth",
     "tags",
 }
+_STEEL_API_URL = "https://api.steel.dev"
+_STEEL_ALLOWED_OPTIONS = {
+    "blockAds",
+    "extensionIds",
+    "proxyUrl",
+    "region",
+    "sessionContext",
+    "solveCaptcha",
+    "stealthConfig",
+    "useProxy",
+    "userAgent",
+}
 DEFAULT_BROWSER_CDP_URL = os.environ.get(
     "CLAWBENCH_BROWSER_CDP_URL",
     "http://127.0.0.1:9222",
@@ -156,6 +168,12 @@ class _KernelApiError(RuntimeError):
         self.status = status
 
 
+class _SteelApiError(RuntimeError):
+    def __init__(self, message: str, *, status: int | None = None) -> None:
+        super().__init__(message)
+        self.status = status
+
+
 def _parse_options(raw: str | None) -> dict[str, Any]:
     if not raw:
         return {}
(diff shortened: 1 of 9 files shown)


Perry2004:
Thanks for the contribution 🎉. 
I'll review this PR as a priority.

src/clawbench/runner/run_support/browser_runtime/providers.py
```diff
@@ -232,22 +263,184 @@ def cleanup(self, session: BrowserSession) -> None:
...
+        method: str,
+        path: str,
+        payload: dict[str, Any] | None = None,
+    ) -> dict[str, Any]:
+        data = (
+            json.dumps(payload, separators=(",", ":")).encode()
+            if payload is not None
+            else None
+        )
+        headers = {"Content-Type": "application/json"}
```
Perry2004: Seems Steel is requiring a non-default UA, otherwise it's giving `HTTP 403, body: "error code: 1010`.

Perry2004 (changes requested):
Please fix the issue mentioned above and also update https://github.com/TIGER-AI-Lab/ClawBench/blob/187cd252bc60af8ac3a2c98a87c9316e49a5ac75/src/clawbench/runner/run.py#L197.

Issue #309, opened 2026-08-19
feat(eval): support DeepSeek Harness and run a reproducible ClawBench benchmark


reacher-z:
## Goal

Add [DeepSeek Harness](https://github.com/deepseek-ai/deepseek-harness) (`dsh`) as a first-class ClawBench harness, then run and publish a reproducible ClawBench evaluation with it.

DeepSeek Harness is an official DeepSeek open-source agent harness built around an “everything is a plugin” architecture. It is currently in developer preview and may introduce breaking changes, so the integration should pin an exact revision/version.

### 1. Harness adapter

- Add a `deepseek-harness` / `dsh` adapter alongside the existing OpenClaw, HermesAgent, and Claude Code paths.
- Provide setup and launch scripts, e.g. `setup-deepseek-harness.sh` and `run-deepseek-harness.sh`.
- Map ClawBench tasks into the dsh session/task interface.
- Capture the standard ClawBench trace bundle:
  - `recording.mp4`
  - `actions.jsonl`
  - `agent-messages.jsonl`
  - `requests.jsonl`
  - `interception.json`
  - `run-meta.json`
- Record the dsh package version or Git commit, plugin set, model configuration, corpus revision, and ClawBench commit in `run-meta.json`.
- Preserve ClawBench’s existing timeout, interception, judge, retry, and failure-taxonomy semantics.

### 2. Reproducibility smoke test

Run a small preregistered subset before the full evaluation:

- 5–10 tasks spanning multiple sites and metaclasses
- One inexpensive supported model/configuration
- Fixed task IDs, corpus commit, dsh revision, judge, limits, and seeds where applicable

Verify that:

- dsh completes tasks end-to-end under ClawBench;
- every attempted task produces a complete or explicitly failed trace bundle;
- failures are attributed to the agent, harness/adapter, site, model/API, or grader infrastructure;
- saved artifacts can be rescored without rerunning the agent.

### 3. Benchmark run

After the smoke test passes:

- Run the full ClawBench V2 corpus with DeepSeek Harness.
- If budget permits, also run V1 for historical comparability.
- Evaluate at least one model already tested under another harness, allowing the comparison to isolate harness effects.
- Match the model endpoint, task snapshot, judge, per-task limits, retry policy, and evaluation window.
- Run multiple trials, or clearly label the result as single-trial; do not silently mix protocols.

Track at minimum:

- Reward / task success rate
- Stage-1 interception rate
- Stage-2 judged success rate
[...]

PR #347, opened 2026-09-09
feat(runner): record code, corpus, and agent revisions in run-meta.json

vaibhavdabas16:
Advances #309 §1. Does not close it — see the scoping note below, which is the part I'd most like a maintainer's read on.

## Why this, and not the `dsh` adapter

#309 §1 asks for two things: a `deepseek-harness` adapter, and *"Record the dsh package version or Git commit, plugin set, model configuration, corpus revision, and ClawBench commit in `run-meta.json`."*

I did the second and deliberately stopped short of the first. Adding a harness here is mechanical — a directory of scripts plus a `harnesses.yaml` entry — but the `setup-*.sh` and `run-*.sh` contents depend entirely on `dsh`'s actual CLI surface, config format, and transcript shape, and DeepSeek Harness is in developer preview. I can't build the image to check, so I'd be writing a plausible-looking `run-deepseek-harness.sh` that would probably be wrong in ways nobody could see from the diff. That seemed worse than not writing it. **Happy to do the adapter as a follow-up if someone can share a working `dsh` invocation, or point me at the preview docs.**

The provenance half turned out to be a real gap, and a general one — it applies to every harness, not just `dsh`.

## The gap

`run-meta.json` records *what* ran — task, model, harness name, image ids — but not the revisions those names resolved to. Two runs of "openclaw on v2" a month apart are indistinguishable in the artifact even when the agent, the corpus, and ClawBench itself have all moved. That makes a published leaderboard row labelled rather than reproducible, and it blocks §2's *"fixed task IDs, corpus commit, dsh revision"* and §4's *"document the exact reproduction commands and environment"* regardless of which harness is being added.

## What lands

A `provenance` block alongside `runtime`:

```json
"provenance": {
  "clawbench_version": "0.10.0",
  "commit": "3f3599d...", "branch": "main", "dirty": false,
  "corpus": {"suite": "v2", "path": "test-cases/v2", "revision": "62ee923..."},
  "harness": {
    "name": "openclaw",
    "image_id": "sha256:...",
    "agent_version": "2026.3.13",
    "pinned_versions": {"openclaw": "2026.3.13"}
  }
}
```

- **`corpus.revision`** is the last commit that touched that suite — two runs sharing it saw the same task text.
- **`harness.pinned_versions`** is parsed from the version pins in the harness Dockerfile (`opencode-ai@1.4.4`, `@playwright/mcp@0.0.70`, `pip install x==1.2`). It reads what the image was *built from* rather than shelling into a running container, so it costs nothing at run time and works for a harness that is already gone. This answers the issue's "package version and plugin set" for any harness. Verified against the bundled Dockerfiles: openclaw, opencode, claude-code, browser-use all resolve; `null` (no pins) correctly reports nothing.

## Failure behaviour, which is most of the design

Every lookup is best-effort and returns `None` rather than raising. A PyPI install has no git checkout; a container host may have no `git`; a task from an explicit `--cases-dir` has a history that isn't ClawBench's to claim. **A missing provenance field must never fail a run that otherwise succeeded.**

One distinction worth flagging: `dirty: null` means *the lookup failed*, which is not the same claim as `false`. Anyone filtering runs by "clean tree" should check `is False`, not truthiness — the cookbook section says so explicitly.

Lookups are `lru_cache`d because a batch builds one block per task and the answers cannot change inside a process.

## Testing

`tests/test_provenance.py` — 16 cases: pin extraction for each bundled harness, that `FROM node:24-slim` and `COPY --from=...uv:0.11.6` are not mistaken for agent pins, pip-style pins, corpus resolution for bundled and external case dirs, and that a missing checkout or missing `git` binary yields nulls rather than an exception. `test_results_and_metadata.py` asserts the block reaches `run-meta.json` and reuses the same image id as `runtime`.

Full suite passes locally (`291 passed, 10 skipped`); `ruff` and `pyright` clean.

### DIFF
--- a/src/clawbench/runner/run_support/provenance.py
+++ b/src/clawbench/runner/run_support/provenance.py
@@ -1,0 +1,198 @@
+"""Which exact code, corpus, and agent produced a run.
+
+`run-meta.json` already records *what* was run — task, model, harness name,
+image ids. It does not record the revisions those names resolved to, so two
+runs of "openclaw on v2" a month apart are indistinguishable in the artifact
+even when the agent, the corpus, and ClawBench itself all moved.
+
+This module fills that gap: the ClawBench version and commit, the corpus
+revision, and the agent versions pinned by the harness image. Every lookup is
+best-effort and returns ``None`` rather than raising — a PyPI install has no
+git repository, a container host may have no `git` at all, and a missing
+provenance field must never fail a run that otherwise succeeded.
+
+Results are cached because a batch run builds one of these per task and the
+answers cannot change inside a single process.
+"""
+
+from __future__ import annotations
+
+import re
+import subprocess
+from functools import lru_cache
+from importlib.metadata import PackageNotFoundError, version
+from pathlib import Path
+from typing import Any
+
+from clawbench.utils.paths import ASSET_ROOT, HARNESS_ROOT, SOURCE_ROOT
+
+_GIT_TIMEOUT_S = 10
+
+# Version pins declared in a harness Dockerfile: `pkg@1.2.3` for npm and
+# `pkg==1.2.3` for pip. This reads the pins the image was built from rather
+# than asking the agent for its version, which would need a running container.
+_NPM_PIN_RE = re.compile(r"(?<![\w@/])((?:@[\w.-]+/)?[\w.-]+)@(\d[\w.+-]*)")
+_PIP_PIN_RE = re.compile(r"([\w.-]+)==(\d[\w.+-]*)")
+# Lines that pin something without installing an agent.
+_PIN_SKIP_RE = re.compile(r"^\s*(#|FROM |COPY |ENV |LABEL )", re.IGNORECASE)
+
+
+def _git(repo: Path, *args: str) -> str | None:
+    try:
+        result = subprocess.run(
+            ["git", "-C", str(repo), *args],
+            capture_output=True,
+            text=True,
+            timeout=_GIT_TIMEOUT_S,
+        )
+    except (OSError, subprocess.SubprocessError):
+        return None
+    if result.returncode != 0:
+        return None
+    return result.stdout.strip() or None
+
+
+@lru_cache(maxsize=1)
+def clawbench_version() -> str | None:
+    try:
+        return version("clawbench-eval")
+    except PackageNotFoundError:
+        return None
+
+
+@lru_cache(maxsize=1)
+def _repo_root() -> Path | None:
+    """The ClawBench git checkout, when running from source rather than a wheel."""
+    if SOURCE_ROOT is None or not (SOURCE_ROOT / ".git").exists():
+        return None
+    return SOURCE_ROOT
+
+
+@lru_cache(maxsize=1)
+def clawbench_commit() -> dict[str, Any]:
+    """Commit, branch, and dirty state of the ClawBench checkout."""
+    repo = _repo_root()
+    if repo is None:
+        return {"commit": None, "branch": None, "dirty": None}
+    status = _git(repo, "status", "--porcelain")
+    return {
+        "commit": _git(repo, "rev-parse", "HEAD"),
+        "branch": _git(repo, "rev-parse", "--abbrev-ref", "HEAD"),
+        # `git status` succeeding with no output means a clean tree; a failed
+        # lookup returns None, which is not the same claim as "clean".
+        "dirty": (status != "") if status is not None else None,
+    }
+
+
+def _relative_to_assets(path: Path) -> str | None:
+    try:
+        return path.resolve().relative_to(ASSET_ROOT.resolve()).as_posix()
+    except (OSError, ValueError):
+        return None
+
+
+@lru_cache(maxsize=8)
+def _corpus_commit(suite_path: str) -> str | None:
+    """Last commit that touched this corpus directory."""
+    repo = _repo_root()
+    if repo is None:
+        return None
+    return _git(repo, "log", "-1", "--format=%H", "--", suite_path)
+
+
+def corpus_meta(task_dir: Path | None) -> dict[str, Any]:
+    """Which corpus a task came from, and at which revision.
+
+    A task outside the bundled corpora (an explicit ``--cases-dir``) reports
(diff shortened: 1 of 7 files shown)


src/clawbench/runner/run_support/provenance.py
```diff
@@ -0,0 +1,198 @@
...
+    repo = _repo_root()
+    if repo is None:
+        return {"commit": None, "branch": None, "dirty": None}
+    status = _git(repo, "status", "--porcelain")
+    return {
+        "commit": _git(repo, "rev-parse", "HEAD"),
+        "branch": _git(repo, "rev-parse", "--abbrev-ref", "HEAD"),
+        # `git status` succeeding with no output means a clean tree; a failed
+        # lookup returns None, which is not the same claim as "clean".
+        "dirty": (status != "") if status is not None else None,
```
Perry2004, line 112: `dirty` would never be `False`. 
Even on a clean checkout, the `_git()` returns None since it converts empty string to `None`.

src/clawbench/runner/run_support/provenance.py
```diff
@@ -0,0 +1,198 @@
...
+    suite_path = "/".join(parts[:2])
+    return {
+        "suite": parts[1] if len(parts) > 1 else parts[0],
+        "path": suite_path,
+        "revision": _corpus_commit(suite_path),
+    }
+
+
+@lru_cache(maxsize=32)
+def harness_pins(harness: str) -> dict[str, str]:
```
Perry2004, line 154: This checks the version pinned in the `Dockerfile` and setup scripts, but they're not necessarily the version being ran due to the existence of `--no-build` flag. 
This should be set to either Unknown of None under such cases.

src/clawbench/runner/run_support/provenance.py
```diff
@@ -0,0 +1,198 @@
...
+
+from clawbench.utils.paths import ASSET_ROOT, HARNESS_ROOT, SOURCE_ROOT
+
+_GIT_TIMEOUT_S = 10
+
+# Version pins declared in a harness Dockerfile: `pkg@1.2.3` for npm and
+# `pkg==1.2.3` for pip. This reads the pins the image was built from rather
+# than asking the agent for its version, which would need a running container.
+_NPM_PIN_RE = re.compile(r"(?<![\w@/])((?:@[\w.-]+/)?[\w.-]+)@(\d[\w.+-]*)")
+_PIP_PIN_RE = re.compile(r"([\w.-]+)==(\d[\w.+-]*)")
```
Perry2004: It is not handling cases with `[]` such as litellm[proxy].

src/clawbench/runner/run_support/provenance.py
```diff
@@ -0,0 +1,198 @@
...
+from typing import Any
+
+from clawbench.utils.paths import ASSET_ROOT, HARNESS_ROOT, SOURCE_ROOT
+
+_GIT_TIMEOUT_S = 10
+
+# Version pins declared in a harness Dockerfile: `pkg@1.2.3` for npm and
+# `pkg==1.2.3` for pip. This reads the pins the image was built from rather
+# than asking the agent for its version, which would need a running container.
+_NPM_PIN_RE = re.compile(r"(?<![\w@/])((?:@[\w.-]+/)?[\w.-]+)@(\d[\w.+-]*)")
```
Perry2004: This only handles @version but not @revisions.

Perry2004 (changes requested):

vaibhavdabas16:
Thanks — all four were real, and the first two were the interesting ones. Fixed in b7bc2e0…78754ab.

**`dirty` could never be `False`.** Correct, and worse than it looked: `_git()` ended in `return result.stdout.strip() or None`, so a clean `git status --porcelain` — which succeeds with *no output* — came back as `None`, identical to a failed lookup. My own test asserted `dirty is False` and passed only because it monkeypatched `_git` to return `""`, bypassing the exact conversion that caused the bug. The mock asserted the intent while the code did the opposite.

`_git()` now returns `None` only when the command could not run, and `""` when it ran and said nothing; callers wanting a non-empty value ask explicitly. The test runs against a real temporary git repo — I verified it fails against the old implementation rather than just passing against the new one.

**`--no-build`.** Also correct. One wrinkle worth flagging: keying off the flag alone would have been wrong in the opposite direction, because `clawbench-batch` builds the image once (`batch.py:734`) and then passes `--no-build` to *every* child. Trusting the flag would have marked the entire batch path — i.e. how leaderboard numbers are actually produced — as unverified.

So the signal is set by `docker_build()` itself, and children inherit it through the environment the same way `CONTAINER_ENGINE` already does. Pins now carry `pins_source`:

| Value | Meaning |
|---|---|
| `"dockerfile"` | image built from this checkout during this run; pins are what ran |
| `"unverified"` | image reused via `--no-build`; `pinned_versions` and `agent_version` are `null` |

A batch run reports `"dockerfile"`; a bare `clawbench-run --no-build` reports `"unverified"`. `image_id` is still reported either way, since that remains a fact about the run.

**Extras.** `litellm[proxy]==1.77.3` matched *nothing* — the pin vanished silently rather than being recorded under a wrong name. Now `{"litellm[proxy]": "1.77.3"}`.

**Revisions.** Now handles `pkg@github:owner/repo#<sha>`, `pkg@git+https://…#<ref>`, and pip's PEP 508 `pkg @ git+https://…@<ref>`. That is precisely how a developer-preview agent gets pinned, so it matters for the `dsh` adapter this PR is groundwork for.

One deliberate exclusion: a floating dist-tag (`pkg@next`, `pkg@latest`) is **not** collected. It names a moving target, so recording it as a pin would be a false claim — `agent_version` is `null` for a harness pinned that way. Happy to reverse that if you'd rather see the tag recorded.

On "and setup scripts": I checked, and no bundled harness pins versions in its `setup-*.sh` — all installs are in the Dockerfiles. So Dockerfile-only parsing is complete today, but it would silently miss a future harness that installs in setup. Say the word if you want the setup scripts scanned too.

CI is green; the full suite passes locally (300 passed, 10 skipped).

Perry2004 (approved):
LGTM🫡
Thanks for the contribution.

PR #368, opened 2026-09-13
fix(rescore): require a matching cache fingerprint, not just match is not None

### LINKED ISSUE #363: [P1] rescore: stale judge caches are silently attributed to a newly requested judge model
## Problem
Both pending selection and `rescore_one` reuse any cached non-null verdict, without comparing judge identity/configuration, rubric revision, instruction or intercepted evidence. The aggregate then unconditionally writes `roll['judge_model'] = args.judge_model`. Requesting a new judge can therefore produce a report labeled with that judge even when every verdict came from an older judge.

Sources: [cache reuse](https://github.com/TIGER-AI-Lab/ClawBench/blob/9dd9d44/src/clawbench/eval/rescore.py#L74), [pending selection](https://github.com/TIGER-AI-Lab/ClawBench/blob/9dd9d44/src/clawbench/eval/rescore.py#L279), [aggregate attribution](https://github.com/TIGER-AI-Lab/ClawBench/blob/9dd9d44/src/clawbench/eval/rescore.py#L352).

## Verified reproduction
Create one intercepted run with `judge.json = {"match": true, "judge_model": "old-judge"}`; run `rescore.main()` with `--judge-model new-judge --rubric strict`, a temporary dummy model configuration, and a mocked judge function.

Observed: exit **0**, judge calls **0**, `rescore-summary.json.judge_model = "new-judge"`, `n_match_strict = 1`. Direct `rescore_one(..., judge_model='new-judge', force=False)` returns the old-judge object unchanged. No network call was made.

## Acceptance criteria
- [ ] Define a non-secret evaluation fingerprint covering judge identity/config, rubric/prompt revision, instruction and intercepted input.
- [ ] Reuse only a valid, matching cache; handle legacy cache entries explicitly (recompute or report unknown provenance).
- [ ] Preserve actual per-verdict provenance and reject or explicitly describe mixed-provenance aggregates.
- [ ] Keep `--force` as an explicit recomputation option; ordinary model changes must not silently mislabel results.
- [ ] Test changing model, input and rubric revision, and unchanged-cache reuse.

Suggested priority P1: misleading benchmark attribution. Related to provenance PR #347, but run metadata alone does not fix this evaluator cache contract. #313 fixes paths/config loading; #319 fixes judge transports; neither addresses cache identity.

@Perry2004 please review the cache/provenance contract and reporting requirements.

AmirF194:
## What does this PR do?

Fixes #363.

`rescore_one`'s cache-hit check only tested whether `match` was set, never whether the cached verdict came from the judge, config, rubric, instruction and intercepted evidence being requested now. Rescoring under a new `--judge-model` replayed the old judge's verdicts with zero live calls, and the batch rollup unconditionally stamped `roll["judge_model"]` with the newly requested judge even though nothing was actually reran. The pending-selection loop in `main()` had the same gap, so a stale run could be excluded from rescoring entirely.

### Fix

- `_cache_fingerprint()` builds a non-secret identity from the judge model, its config (`api_key`/`api_keys` excluded), the rubric, the instruction and the intercepted evidence, stored as `cache_fingerprint` alongside each verdict.
- `_cache_is_stale()` is the single place that decides whether a cached verdict answers the current request. Both `rescore_one` and the pending-selection loop call it, so they can't disagree.
- A verdict with no `cache_fingerprint` (everything written before this change) is treated as unknown provenance and recomputed, not trusted just because `match` is set.
- `aggregate_batch()` now tracks which judge produced each verdict. The rollup sets `judge_model_mixed: true` and prints a warning if a batch still mixes judges (e.g. after `--limit`), instead of silently reporting one.
- `--force` is unchanged.

Two notes on scope: the fingerprint doesn't version the judge prompt text itself, only the inputs `rescore_one` receives, so an in-place prompt edit wouldn't invalidate old caches. And "rubric revision" from the issue is already handled structurally, since `strict`/`lenient` write separate files and never share a cache entry; the fingerprint includes `rubric` anyway for defense in depth.

## Corpus

- [ ] v2
- [ ] v1
- [ ] both
- [x] not applicable

<sub>Host-side evaluator change; no task data involved.</sub>

## Test plan

- [x] `tests/test_rescore_cache_provenance.py`, new: the issue's own repro (cache from an old judge is not reused), a legacy cache with no fingerprint, an instruction change, a judge-config change, api keys excluded from the fingerprint, a matching cache still reused, and `aggregate_batch` flagging mixed judge provenance.
- [x] `tests/test_rescore_retarget.py`, updated: its "cached scored verdict is not retried" case now carries a matching fingerprint, since an unqualified cache is no longer trusted.
- [x] Full suite in a clean `python:3.11-slim` container via `uv run --frozen pytest`: 293 passed, 2 skipped, and every new/changed test fails without this fix. `ruff check`, `ruff format --check`, `pyright`, `uv build` + `twine check`, the three console-script `--help` checks, and the Markdown link checker (triggered by the CHANGELOG edit): all clean.

I have not exercised a real judge-provider call here; the tests drive `rescore_one` through fake judge functions, not live API traffic.

## Related issues

Fixes #363.


AmirF194:
CI's been green for a few days with no review yet. The fix is scoped to rescore.py's cache-fingerprint check plus the two new regression tests, so it should be a quick look if you get to it.

